PR #103, opened 2015-01-07
In Mutex, call MutexFactory wrapper functions instead of directly callin...

jelu:
...g the function pointers so the MutexFactory enabled bool has effect.

This was detected when testing initializing with mutex functions and then reinitializing without them and they where still called. V1 destroys the MutexFactory instance on finalize so its recreated on the next initialize, V2 does not do anything with the MutexFactory on finalize so the next initialize just disable it but since the Mutex called the functions pointers directly it did not have the desired effect. Maybe the Mutex class shouldn't be a friend to the MutexFactory class.

### DIFF
--- a/src/lib/common/MutexFactory.cpp
+++ b/src/lib/common/MutexFactory.cpp
@@ -43,7 +43,7 @@
 // Constructor
 Mutex::Mutex()
 {
-	isValid = (MutexFactory::i()->createMutex(&handle) == CKR_OK);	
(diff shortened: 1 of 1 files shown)


jschlyter:
@bellgrim, please review

bellgrim:
#build

bellgrim:
Thanks, the changes are looking good and can be merged.

PR #105, opened 2015-01-22
Do not install cleanup hook in shared library

kindkaktus:
atexit() handler it called when an app exits and NOT when a library is unloaded.
Because the app most likely unloads the library before calling atexit() handler, we end up calling a function from the already unloaded lib, which will result in crash.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -468,8 +468,14 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
 	// Set the state to initialised
 	isInitialised = true;
 
-	// Hook cleanup on dlclose() or exit()
+#ifndef CRYPTOKI_EXPORTS
+	// Install cleanup hook on app exit() unless we are shared library
+
+	// atexit() handler it called when an app exits and NOT when a library is unloaded.
+	// Because the app most likely unloads the library before calling atexit() handler, 
+	// we end up calling a function from the already unloaded lib, which will result in crash.
 	atexit(libcleanup);
+#endif
 
 	return CKR_OK;
 }


bellgrim:
#build

bellgrim:
I found this email conversation when looking at the history of this code:
{{{
hook a call to SoftHSM::i()->C_Finalize(NULL) before DSO unloading (__attribute((destructor))) and exit (atexit()). BTW if you want to handle only OSes where atexit() is called by dlclose() (Linux, Windows, Solaris) it is pretty easy to do (just add a call to atexit() at the end of C_Initialize())
}}}

May we do want to call atexit() but just for some OS?

kindkaktus:
> May we do want to call atexit() but just for some OS?

1. First, the crash is reproducible for me on Linux (CentOS-6.6) and on BSD (OpenBSD-5.3), with the evidence on trying to call DSO function after the DSO gets unloaded. 
2. Second, if you still want to schedule C_Finalize() on DSO unload, keep in mind that it will be platform-specific. For *nix systems with gcc this will likely be __attribute((destructor)). For Windows with MSVC you shall hook in DllMain().

fxdupont:
I am in favour to reject this pull request:
 - I don't believe most applications unload DSOs.
 - many systems (Linux since glib 2.2.3, FreeBSD, Windows (**)) call the atexit handlers when they point to the DSO address space at unloading (dlclose, etc) time. The idea is to offer a way to cleanup the DSO, i.e., exactly what we want.
 - only AIX provides an unatexit function to deregister handlers
in fact the proposed pull request just exchanges a bug for another (and already fixed)...

I suggest:
 - to make the call to atexit conditional inside a #ifdef #endif
 - to create a ticket for a m4 configuration file to check if the atexit handler is called at unloading time
 - to add a new configuration option for systems where this behaviour was not yet added to the C/C++ runtime

(**) https://msdn.microsoft.com/en-us/library/988ye33t.aspx

jschlyter:
ping @bellgrim

bellgrim:
Would something like this work better?

If yes, then I have a patch ready including autoconf.

```
diff --git a/src/lib/SoftHSM.cpp b/src/lib/SoftHSM.cpp
index 95fda61..3053e1c 100644
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -264,6 +264,9 @@ static CK_ATTRIBUTE bsAttribute(CK_ATTRIBUTE_TYPE type, const ByteString &value)
 }
 #endif
 
+#ifdef HAVE_FUNC_ATTRIBUTE_DESTRUCTOR
+__attribute__((__destructor__))
+#endif
 static void libcleanup()
 {
        SoftHSM::i()->C_Finalize(NULL);
@@ -473,8 +476,10 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
        // Set the state to initialised
        isInitialised = true;
 
+#ifndef HAVE_FUNC_ATTRIBUTE_DESTRUCTOR
        // Hook cleanup on dlclose() or exit()
        atexit(libcleanup);
+#endif
 
        return CKR_OK;
 }
```

bellgrim:
The code is in #124

bellgrim:
@kindkaktus Does it work better with #124 ?

kindkaktus:
@bellgrim Yes, the patch solves the issue.

bellgrim:
Thank you, closing this issue.

PR #112, opened 2015-03-14
SUPPORT-157: Derived secrets were truncated after encryption and could t...

bellgrim:
...hus not be decrypted.


bellgrim:
#build

bellgrim:
Jakob, please check the changes and then merge it if they are ok.

PR #125, opened 2015-06-01
Add configuration file for Travis CI

jschlyter:


jschlyter:
Now testing both Botan & OpenSSL as well as file & sqlite object stores. Example at https://travis-ci.org/jschlyter/SoftHSMv2.

.travis.sh
```diff
@@ -0,0 +1,35 @@
...
+	CONF_CRYPTO="$CONF_CRYPTO --with-crypto-backend=openssl --with-openssl=/usr"
+	CONF_CRYPTO="$CONF_CRYPTO --disable-gost"
+	openssl version -a
+	;;
+*)
+	echo "Unknown crypto backend"
+	exit 1
+esac
+
+case $OBJSTORE in
```
bellgrim, line 21: Maybe we do not need to build with and without "--with-objectstore-backend-db". The file backend will always be supported. It is not like OpenSSL or Botan, where you only have one backend in the build.

Perhaps we always can make the test build with "--with-objectstore-backend-db --with-migrate"? Or do we want to test a build without SQLite support?

.travis.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+CONF_CRYPTO=""
+CONF_OBJSTORE=""
+
+case $CRYPTO in
+botan)
+	CONF_CRYPTO="$CONF_CRYPTO --with-crypto-backend=botan --with-botan=/usr"
+	CONF_CRYPTO="$CONF_CRYPTO --disable-ecc --disable-gost"
```
bellgrim, line 9: Too bad that Travis is using Ubuntu 12.04, which only include Botan 1.10.0. Thus not supporting ECC and GOST.

jschlyter, line 9: Yes. On the other hand, OpenSSL has ECC so that part is covered.

.travis.sh
jschlyter, line 21: Yes, without SQLite would be useful I believe.

bellgrim:
One more question before merge... I guess that we do not want these files included in the dist?

jschlyter:
Correct, this is repo only.

PR #128, opened 2015-08-11
Fix for bug in C_GetAttributeValue, which bails at the first error

NWilson:
The PKCS spec requires processing to fill in all attributes in the array, if one of the benign error codes (SENSITIVE, TYPE_INVALID, BUFFER_TOO_SMALL) is returned.


bellgrim:
#build

bellgrim:
Thank you, merging this patch.

PR #154, opened 2015-09-19
softhsm2-util crash under WIndows

agrandville:
MutexFactory is destroyed before SecureMemoryRegistry but SecureMemoryRegistry needs MutexFactory so destruction of SecureMemoryRegistry make the mess !
I propose to invert instanciation sequence of  SecureMemoryRegistry and MutexFactory to allow the atexit function calling the destructors into the right order. 
The second commit is about warnings and compilations options under windows

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -69,6 +69,22 @@
 
 #include <stdlib.h>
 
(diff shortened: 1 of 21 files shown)


bellgrim:
#build

bellgrim:
Thank you for the patches, they are looking good!

Issue #162, opened 2015-10-16
A marked as CA certificate cannot be written in a softhsmv2 db


nmav:
Trying to write a certificate in a softhsm db with CKA_CERTIFICATE_CATEGORY fails with: CKR_ATTRIBUTE_READ_ONLY

How reproducible:
1. cat >config
directories.tokendir = db
objectstore.backend = file

2. export SOFTHSM2_CONF=config
3. mkdir db
4. softhsm2-util --init-token --slot 0 --label test --so-pin 1234 --pin 1234
5. p11tool --provider /usr/lib64/pkcs11/libsofthsm2.so --write --mark-ca --load-certificate any-cert.pem  --label test --so-login

Output:
Error writing certificate: PKCS #11 error in attribute

Expected Output:
Success.

Writing the same certificate without the mark-ca flag works fine.

The PKCS#11 Spy output is at:
https://bugzilla.redhat.com/show_bug.cgi?id=1272453

bellgrim:
I tried to replicate this using GnuTLS 3.4.5, but it fails to create the object in the correct way. It tries to C_CreateObject() when logged in as SO. It does not set CKA_PRIVATE (which defaults to true). A SO cannot create private objects.

bellgrim:
#164 have a fix for this.

Also had to add --no-mark-private to the p11tool command to get this to work.

nmav:
The login as SO is because of the --so-login option. You may switch it to simply --login.

bellgrim:
Fixed in #164

PR #164, opened 2015-10-21
Not possible to create certificate objects

bellgrim:
This will fix issue #162.

CKA_CERTIFICATE_CATEGORY, CKA_NAME_HASH_ALGORITHM, and CKA_JAVA_MIDP_SECURITY_DOMAIN have invalid checks (should not be != but ==). Also removing the checks because they are handled when ck8 is not set on these attributes.

While at it, removing other abundant checks. This is handled by the parent function.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,9 @@ SoftHSM develop
   (Thanks to Jaroslav Imrich)
 
 Bugfixes:
+* Issue #162: Not possible to create certificate objects containing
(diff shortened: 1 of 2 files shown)


bellgrim:
#build

jschlyter:
FWIW, Travis does not need the #build thingy, it will build automagically.

bellgrim:
It is so that Jenkins will build and verify the code.

bellgrim:
#build again. The build was aborted in Jenkins.

PR #172, opened 2015-11-26
Support for MD2/RC2/RC4

lhoward:
Lightly tested support for MD2/RC2/RC4, needed for Kerberos interoperability.


lhoward:
address review comments

bellgrim:
Thanks for the pull request. It is looking good, just need some minor tweaks according to my comments above.

Some additional questions:
* There is no CKM_MD4 mechanism in PKCS#11. Do we need MD4 support in the crypto backend?
* We can always include these algorithms, but aren't they deprecated in most applications?

lhoward:
CKM_MD4 would be nice, but it seems like there’s no assigned identifier for it. Because the set of supported mechanisms may vary at runtime, our consumer (PKCS#11 backend for Heimdal) now tests and dynamically falls back to an internal implementation for unsupported mechanisms.

Regarding them being deprecated, yes, although they may still be in used for Windows interoperability in Kerberos. My goal was mostly to make the Heimdal test suite pass.

> On 27 Nov 2015, at 11:27 PM, Rickard Bellgrim <notifications@github.com> wrote:
> 
> Thanks for the pull request. It is looking good, just need some minor tweaks according to my comments above.
> 
> Some additional questions:
> 
> There is no CKM_MD4 mechanism in PKCS#11. Do we need MD4 support in the crypto backend?
> We can always include these algorithms, but aren't they deprecated in most applications?
> —
> Reply to this email directly or view it on GitHub <https://github.com/opendnssec/SoftHSMv2/pull/172#issuecomment-160129600>.
> 

--
www.lukehoward.com
soundcloud.com/lukehoward

bellgrim:
> Not sure I understand this (will need to read spec) – can’t find any other references to CKA_VALUE_LEN in P11Objects.cpp.

P11Objects.cpp use the SoftHSM class called P11AttrValueLen for CKA_VALUE_LEN. There are some comments in P11Attributes.cpp that link it to the PKCS#11 attribute.

We have not yet upgrade to PKCS#11 v2.40 so I use this as a reference, page 22 in:
ftp://ftp.rsasecurity.com/pub/pkcs/pkcs-11/v2-30/pkcs-11v2-30m2-d3.pdf

bellgrim:
The build is failing, MD2 is not enabled for OpenSSL on Debian. Will need some configure checks, and disable MD2 if it is not supported.

vanrein:
Please take note of
  * RFC 6149 - MD2 to Historic Status
  * RFC 6649 - Deprecate DES, RC4-HMAC-EXP, and Other Weak Cryptographic Algorithms in Kerberos
  * RFC 7465 - Prohibiting RC4 Cipher Suites

The algorithms proposed here are not only ancient, they are actively deprecated and advised against.  Meaning, they should not be included in new products.  Some of the algorithms can probably be reversed, so there is no added value to putting it in an HSM -- other than the false feeling of security due to a graphical indication of being FIPS-compliant perhaps.

I've seen that many people think that Kerberos5 is still using old encryption types; it is not, its encryption type tagging allows the usual agreement on (modern) ciphers.  The only thing holding this back is the human interaction involved in setting up new passwords that bootstrap the daily SSO session.  But passwords that still use this crypto go so long back that they could use a refreshment; the passwords were probably picked when we were less aware of the dangers of using the names of our spouses, cats or first-borns.  Such passwords could use an update anyway, because all the traffic deriving from it can be decrypted after the secrets have been guessed (no PFS due to symmetric crypto).

Kerberos5 has long moved beyond old-fashioned ciphers, and is nowadays relying on things like aes256-cts-hmac-sha1-96.  FWIW, I recently demonstrated that [AES-CTS can be emulated](https://github.com/arpa2/kerberos-pkcs11) on PKCS #11.

Also note that there is no CKM_ method for C_UnwrapKey() that interprets Kerberos5 messages in today's PKCS #11 so this would only work for the initial password used for start-of-the-daily-session login (AS exchange).  When generating new secrets it's possible to take the opportunity to also move to better crypto, so the only thing that the added mechanisms add for Kerberos5 would be to import pre-existing passwords to weak ciphers.  Let me repeat: once you've got this secret decrypted you can decrypt all other traffic derived from it, due to symmetric key crypto and currently no PFS support.

In conclusion, I don't believe this adds valuable substance to SoftHSMv2 or to Kerberos; it may however do harm to both technologies.  I don't believe the World is looking for more ways to acces known-to-be-weak algorithms, because they might be helpful in downgrade attacks.

@bellgrim - I would flag these facilities as contraversial and ask that they are not included before discussion on the developer's list.  If I have a vote, I will make it say to not include these weak ciphers and risk them taking down the quality level of SoftHSMv2 as well as Kerberos5.

@lhoward - as an independent cryptographer with detailed knowledge of both PKCS #11 and Kerberos5, I suggest you re-iterate your in-house security instead of asking the rest of the World to be compatible with your current, ill-advised cipher preferences.  I hope the above explains clearly enough why.

lhoward:
> On 11 Dec 2015, at 9:16 AM, vanrein <notifications@github.com> wrote:
> 
> @lhoward <https://github.com/lhoward> - as an independent cryptographer with detailed knowledge of both PKCS #11 <https://github.com/opendnssec/SoftHSMv2/pull/11> and Kerberos5, I suggest you re-iterate your in-house security instead of asking the rest of the World to be compatible with your current, ill-advised cipher preferences. I hope the above explains clearly enough why.
> 

I was implementing PKCS#11 support in Heimdal’s hcrypto for a customer, and although Solaris was the target platform decided to use SoftHSM in order to continue developing on OS X.

I added support for these deprecated ciphers for two reasons:

* To make the test suites pass
* In the real world, interoperability (e.g. with Windows) occasionally trumps security

The patches: take them or leave them. I’ve been working on Heimdal (and MIT for a little less long) for the past 18 years and I really do not care to be lectured to.

— Luke

Issue #171, opened 2015-11-10
Problem while closing library that is initialized but improperly finalized.


halderen:
There is a segmentation violation when closing the softhsm2 library when it was opened dynamically using dlopen().  It happens when the library is initialized using C_Initialize(), but no C_Finalize is called prior unloading the library.

```c
int
main(int argc, char** argv)
{
    int i;
    void* handle;
    void* symbol;
    CK_FUNCTION_LIST_PTR table;
    CK_SESSION_HANDLE session;
    handle = dlopen("/usr/local/lib/softhsm/libsofthsm2.so", RTLD_NOW | RTLD_LOCAL);
    symbol = dlsym(handle, "C_GetFunctionList");
    ((CK_C_GetFunctionList) symbol)(&table);
    table->C_Initialize(&InitArgs);
    return 0;
}
```
Note the dlclose is implicit here, but an explicit dlclose as the end results in about the same.  It will result in
```
Program terminated with signal 11, Segmentation fault.
#0  0x00007f7cc62ad943 in SoftHSM::C_Finalize (this=0x11d5390, pReserved=0x0)
    at SoftHSM.cpp:513
513		if (handleManager != NULL) delete handleManager;
```
Simply because all the vptr is already gone.  My personal take would be to remove the libcleanup() function alltogether, it is an attempt to close off properly when the calling program has not done so (which is the case above), but that won't always work properly.

bellgrim:
Thank you for the report, this has now been fixed in #171

PR #175, opened 2015-12-10
Issue #171: Problem while closing library that is initialized but imp…

bellgrim:
…roperly finalized.

Just use auto_ptr / unique_ptr. The destruction order is the reverse of the construction order. libcleanup() tried to fix this, but it was called to late. Now the destruction order is made according to the dependencies between the classes.


bellgrim:
#build

bellgrim:
#build

bellgrim:
#build

PR #180, opened 2015-12-29
configure script improvements

agrandville:
* autodetect PlatformToolset version from visual studio install folder name
* check if script is launch from a vs command prompt
* add with-toolset param to configure script to overwrite platformtoolset auto detection
* add PLATFORMTOOLSET tokens to the visual studio project files templates

### DIFF
--- a/win32/Configure.py
+++ b/win32/Configure.py
@@ -57,7 +57,8 @@ varnames = ["CUINCPATH",
             "LIBNAME",
             "LIBPATH",
             "PLATFORM",
-            "PLATFORMDIR"]
+            "PLATFORMDIR",
+            "PLATFORMTOOLSET"]
 
 # conditions to stack
 
@@ -89,7 +90,8 @@ withlist = ["botan",
             "crypto-backend",
             "debug-botan",
             "debug-openssl",
-            "openssl"]
+            "openssl",
+            "toolset"]
 
 # general commands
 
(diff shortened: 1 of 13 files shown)


bellgrim:
Thank you for the updated configure script. Could you also make the corresponding changed to Configure.pl?

agrandville:
I'm not sure it's a good idea to maintain simultaneously two methods. Don't you prefere to remove the perl version ?
I'm evaluating a migration to gyp (**a cross platform make project**), if necessary, have a look to [github nodejs project](https://github.com/nodejs/node) to see it in action. What do you think about that ?

Arnaud

bellgrim:
Yes, I think we could remove the perl version. We then also need to update https://github.com/opendnssec/SoftHSMv2/blob/develop/WIN32-NOTES.md

Could you make these changes in this pull request?

A cross plattform make project would be nice. However, it would be a large change that perhaps could take some time to get it up and running. Have never used gyp before, so do not know the pros/cons.

rijswijk:
Rickard Bellgrim wrote:
> Yes, I think we could remove the perl version. We then also need to
> update https://github.com/opendnssec/SoftHSMv2/blob/develop/WIN32-NOTES.md
> 
> Could you make these changes in this pull request?
> 
> A cross plattform make project would be nice. However, it would be a
> large change that perhaps could take some time to get it up and running.
> Have never used gyp before, so do not know the pros/cons.

I'm a bit wary of moving to yet another reinvention of a build
environment, particularly because package maintainers for the various
Linux distros rely on us to stick to something somewhat standardised. If
anything, it might be worthwhile exploring CMake, which will also work
on Windows I think. I've never done anything in anger with CMake though,
but it is pretty well supported on most distros we currently have
packagers for.

Cheers,

Roland

-- 
-- SURFnet bv

Issue #181, opened 2016-01-07
The OSSLCryptoFactory class destructor should not clean up the CRYPTO library state


mtrojnar:
This crashes other code that also uses libcrypto.  See https://github.com/OpenSC/libp11/issues/51#issuecomment-169831003 for details.

bellgrim:
Is there a way of checking if OpenSSL has already been initialized by the calling application?

Compare https://github.com/opendnssec/SoftHSMv2/blob/63f5e89393b12062e0db7ef6429a160f98962e23/src/lib/crypto/OSSLCryptoFactory.cpp#L110 with  https://github.com/opendnssec/SoftHSMv2/blob/63f5e89393b12062e0db7ef6429a160f98962e23/src/lib/crypto/BotanCryptoFactory.cpp#L64

bellgrim:
The order should be:
* Application initialize crypto library
* Application initialize PKCS#11 library
* PKCS#11 library initialize crypto library, if not initialized by the calling application
* Application finalize PKCS#11 library
* PKCS#11 library clean up crypto library, if it was initialized by the PKCS#11 library
* Application clean up crypto library

mtrojnar:
It is a bit more complex.  There is no single call to initialize libcrypto:
https://wiki.openssl.org/index.php/Library_Initialization#libcrypto_Initialization
The application may only initialize a few algorithms it intends to use.  This is safe with libcrypto, because libcrypto initialization (unlike libcrypto clean-up or PKCS#11 initialization) does __not__ remove any accumulated internal state of the library.  The initialization of every libcrypto component (for example an algorithm) has an internal check to prevent double initialization.  It is also perfectly okay for the application to initialize the PKCS#11 library, and then initialize the libcrypto for the algorithms it wants to use directly.

As long as the application is still allowed to use libcrypto after PKCS#11 finalization, the __only__ solution is to leave all the libcrypto clean-up to the application.

jariq:
It might be interesting to check how is PKCS#11 module provided by OpenSC (opensc-pkcs11.so) dealing with this. AFAIK it uses OpenSSL for general cryptographic tasks too so its authors were probably facing the same problem.

mtrojnar:
OpenSC simply never invokes `ENGINE_cleanup()`...

mtrojnar:
This is a critical issue with a very simple fix available as https://github.com/opendnssec/SoftHSMv2/pull/184.  Please merge it.

[...]

PR #184, opened 2016-02-01
Stop discarding the global OpenSSL libcrypto state

mtrojnar:
A single instance of the OpenSSL libcrypto library can be shared by the main application and its libraries, including `libsofthsm2.so`.  To enable it, the libcrypto API checks if each individual subsystem is already initialized, and prevents double initialization.  Initializing the same subsystem by multiple components (the main application or its libraries) is thus safe.

On the other hand, discarding (cleaning up) the state of a libcrypto subsystem breaks all other components using it.  Consequently, only the main application may safely perform cleanup of the OpenSSL libcrypto library.

See https://wiki.openssl.org/index.php/Library_Initialization#libcrypto_Initialization for details.

For example, the existing code causes applications to crash on `ENGINE_cleanup()`, since `libsofthsm2.so` has already executed this function.  See OpenSC/libp11#51 for details.  The code is also likely to cause other, less obvious defects.

This PR fixes opendnssec/SoftHSMv2#181.

### DIFF
--- a/src/lib/crypto/OSSLCryptoFactory.cpp
+++ b/src/lib/crypto/OSSLCryptoFactory.cpp
@@ -204,19 +204,10 @@ OSSLCryptoFactory::~OSSLCryptoFactory()
 		ENGINE_free(eg);
 		eg = NULL;
 	}
-
-	ENGINE_cleanup();
 #endif
 
 	// Destroy the one-and-only RNG
 	delete rng;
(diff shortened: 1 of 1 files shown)


mouse07410:
I confirm that this fix is required, and that it works. Please merge.

bellgrim:
#build

PR #192, opened 2016-02-26
New property in softhsm2.conf; the number of uninitialized tokens tha…

larssilven:
In our project a thread may need to create a new tokens at any time.
Other threads might have open sessions so it is not an alternative to do any "C_Finalize C_Initialize".
We would appreciate if you will accept this request so that we don't have to manage our own repository that we have to keep updated with yours.

Best Regards,
Lars


bellgrim:
If it is within a single application, then all threads should be able to use this new token. If it is between two different application, then currently you have to restart the application using SoftHSM when adding a new token. This is just from right off the top of my head.

This patch could be applied to SoftHSM, but your application would not work with HSM appliances.

From PKCS#11:
> The C_InitToken function may not be sufficient to properly initialize complex tokens. In these situations, an initialization mechanism outside the scope of Cryptoki must be employed. The definition of “complex token” is product specific.

larssilven:
I meant within the same application so it is great if you will accept the patch.

But it was good that you pointed out that an application will not get access to tokens created by other applications afters it has started. Is there any possibility to change this behavior? We have some idea of having a cluster of HSMs that all are using the same tokens. In this case we need to have a mechanism to reload tokens in a running application.

bellgrim:
I checked the code and a new empty slot/token is only added upon start up. A better way would be to always add a new empty token if the old one was initialized. You would thus alternate between C_InitToken and C_GetSlotList/C_GetSlotInfo when adding many new tokens. The current patch will only change that fixed single slot upon start up to a configurable number.

Since you are working within a single application, then we currently do not need to add code for checking updates in the slot list. But it would be a good feature to be able to dynamically adding slots, externally to the application. This would be similar to how new objects can be created in one application and then used by another application.

I do not have the details of your use case, but I do want to point out that you will not get a higher security level by using multiple SoftHSM tokens just within a single application. One case of using multiple SoftHSM tokens is e.g. when a master application is controlling other applications and you want to separate the key stores between them. There are more possibilities with an HSM appliance where you could use e.g. different users, policies and offline tokens.

jariq:
> A better way would be to always add a new empty token if the old one was initialized. You would thus alternate between C_InitToken and C_GetSlotList/C_GetSlotInfo when adding many new tokens.

I second that!

larssilven:
In my opinion it is better to have a fixed slots during the lifetime of an application; C_GetSlotList should always return the same when called from one application.
The reason is that most existing p11 implementations has fixed slots (but sometimes tokens could be inserted and removed from these slots) and hence most p11 implementations are expecting that C_GetSlotList always returns the same list:
At start of your application you call C_GetSlotList to get all existing slots. Then you got iterate through this list to find out if the tokens that you need are present and if not create them if there are enough uninitialized token.

To further try to convince you here follows a description on how it works in other p11 implementations:
Most p11 implementations has a fixed number of slots that can not be changed due to HW restrictions.
But for some you specify how many slots with tokens (uninitialized from start) it should be before starting any application that will use the HSM. Often you then set a high number (maybe 100) so you know that there will always be enough tokens. If the needed data for each uninitialized token is low this is not a problem. Example of such popular HSMs that can be configured for an arbitrary number of slots are: CryptoServer from Utimaco, ProtectServer from SafeNet .

But if you stick to the idea to add a new slot whenever a token is initialized we will try to adapt to this solution instead.

larssilven:
> This would be similar to how new objects can be created in one application and then used by another application.
Is this possible now? Would be great if it is.

bellgrim:
> In my opinion it is better to have a fixed slots during the lifetime of an application; C_GetSlotList should always return the same when called from one application.
The reason is that most existing p11 implementations has fixed slots (but sometimes tokens could be inserted and removed from these slots) and hence most p11 implementations are expecting that C_GetSlotList always returns the same list:

What if you e.g. add a smart card reader when the application is running. Would the list of slots change? Do not know the behavior of e.g. OpenSC.

> At start of your application you call C_GetSlotList to get all existing slots. Then you got iterate through this list to find out if the tokens that you need are present and if not create them if there are enough uninitialized token.

Is there a big difference between having the spare tokens initialized or uninitialized?

> To further try to convince you here follows a description on how it works in other p11 implementations:
Most p11 implementations has a fixed number of slots that can not be changed due to HW restrictions.
But for some you specify how many slots with tokens (uninitialized from start) it should be before starting any application that will use the HSM. Often you then set a high number (maybe 100) so you know that there will always be enough tokens. If the needed data for each uninitialized token is low this is not a problem. Example of such popular HSMs that can be configured for an arbitrary number of slots are: CryptoServer from Utimaco, ProtectServer from SafeNet .

I only have experience of SafeNet Luna SA5 and G5. In those cases, the initialization process is separated from the application using them.

With SoftHSM, uninitialized tokens are currently not stored on disc. The token is only created upon initialization. So the N-number of uninitialized tokens in application X are not same as the ones in application Y. E.g. initializing the spare tokens in both applications will result in 2*N tokens.

> But if you stick to the idea to add a new slot whenever a token is initialized we will try to adapt to this solution instead.

Sorry for the lengthy discussion. Just want to be clear about the implementation and use case. It maybe that we have to change the behavior and handling of slots/tokens based on this discussion.

>> This would be similar to how new objects can be created in one application and then used by another application.

> Is this possible now? Would be great if it is.

Only the object synchronization is available. Adding new tokens currently requires C_Finalize/C_Initialize for the second application to notice the new token. But we could potentially use similar code for signaling changes with the tokens.

rijswijk:
Rickard Bellgrim wrote:
>     In my opinion it is better to have a fixed slots during the lifetime
>     of an application; C_GetSlotList should always return the same when
>     called from one application.
>     The reason is that most existing p11 implementations has fixed slots
>     (but sometimes tokens could be inserted and removed from these
>     slots) and hence most p11 implementations are expecting that
>     C_GetSlotList always returns the same list:
> 
> What if you e.g. add a smart card reader when the application is
> running. Would the list of slots change? Do not know the behavior of
> e.g. OpenSC.

Having worked for a company that produced PKCS #11 libraries for smart
cards, I can comment on that. The answer is: this is implementation
dependent. In the product I worked on we supported dynamic adding and
removal of smart card readers as PKCS #11 slots. Whether or not this was
then disclosed to the application calling the PKCS #11 library depended
on the application. Much of our time was spent writing work arounds for
broken applications that misinterpreted the PKCS #11 standard, so
dynamic addition/removal of slots when a smart card reader was plugged
in or removed was just one more "hack".

To be clear, a smart card PKCS #11 library will normally report a smart
card reader as a slot which, depending on whether a card is inserted or
not will report if a token is present. Whether or not that token is then
initialized depends on the state of the smart card. So in PKCS #11,
slot, token and the status of a token are independent.

>     At start of your application you call C_GetSlotList to get all
>     existing slots. Then you got iterate through this list to find out
>     if the tokens that you need are present and if not create them if
>     there are enough uninitialized token.
> 
> Is there a big difference between having the spare tokens initialized or
> uninitialized?
> 
>     To further try to convince you here follows a description on how it
>     works in other p11 implementations:
>     Most p11 implementations has a fixed number of slots that can not be
>     changed due to HW restrictions.
>     But for some you specify how many slots with tokens (uninitialized
>     from start) it should be before starting any application that will
>     use the HSM. Often you then set a high number (maybe 100) so you
>     know that there will always be enough tokens. If the needed data for
>     each uninitialized token is low this is not a problem. Example of
>     such popular HSMs that can be configured for an arbitrary number of
>     slots are: CryptoServer from Utimaco, ProtectServer from SafeNet .
> 
> I only have experience of SafeNet Luna SA5 and G5. In those cases, the
> initialization process is separated from the application using them.
> 
> With SoftHSM, uninitialized tokens are currently not stored on disc. The
> token is only created upon initialization. So the N-number of
> uninitialized tokens in application X are not same as the ones in
> application Y. E.g. initializing the spare tokens in both applications
> will result in 2*N tokens.

The description above about "most P11 implementations" only holds water
for HSM PKCS #11 implementations. Those for smart cards, USB tokens and
other removal media show markedly different behaviour. In my opinion, we
should not limit SoftHSM v2 to one type of behaviour, so if there is a
need for slots with uninitialised tokens then this should be
configurable behaviour. AFAIK, the original idea was to always have one
slot with an uninitialised token present in SoftHSM v2, and creating a
new token is then simply a matter of initialising that token.

>     But if you stick to the idea to add a new slot whenever a token is
>     initialized we will try to adapt to this solution instead.
> 
> Sorry for the lengthy discussion. Just want to be clear about the
> implementation and use case. It maybe that we have to change the
> behavior and handling of slots/tokens based on this discussion.
> 
>         This would be similar to how new objects can be created in one
>         application and then used by another application.
> 
>     Is this possible now? Would be great if it is.
> 
> Only the object synchronization is available. Adding new tokens
> currently requires C_Finalize/C_Initialize for the second application to
> notice the new token. But we could potentially use similar code for
> signaling changes with the tokens.

If I remember correctly, the way we implemented this in the P11 library
for smart cards I've worked on in the past was that it still required a
C_Finalize/C_Initialize. This is because, if I'm not mistaken, the PKCS
#11 standard mandates that the list of slots is immutable after a call
to C_Initialize.

Hope this additional info helps.

Cheers,

Roland

-- 
-- SURFnet bv

bellgrim:
> If I remember correctly, the way we implemented this in the P11 library
for smart cards I've worked on in the past was that it still required a
C_Finalize/C_Initialize. This is because, if I'm not mistaken, the PKCS
#11 standard mandates that the list of slots is immutable after a call
to C_Initialize.

PKCS#11 v2.20 says that the slot list can be dynamic:

> Furthermore, the set of slots accessible through a Cryptoki library is checked at the time that C_GetSlotList, for list length prediction (NULL pSlotList argument) is called. If an application calls C_GetSlotList with a non-NULL pSlotList, and then the user adds or removes a hardware device, the changed slot list will only be visible and effective if C_GetSlotList is called again with NULL.

But it also says that you can have a static list that is only updated after C_Initialize:

> Even if C_ GetSlotList is successfully called this way, it may or may not be the case that the changed slot list will be successfully recognized depending on the library implementation. On some platforms, or earlier PKCS11 compliant libraries, it may be necessary to successfully call C_Initialize or to restart the entire system.

larssilven:
Most important for us is that we don't have to call C_Finalize/C_Initialize . This since other threads that are using tokens should be able to continue with their work without having to renew their sessions and slot IDs. I think this may be beneficial even for others.
Also I now don't have any problem with that a new slot with an initialized token is added each time the current uninitialized token is initialized. Calling C_ GetSlotList after an initialization will be fine for us.
Maybe it was just in my head that it was stated that the slot list should be fixed. Bellgrims reference that it don't have to be fixed.
I will try to come up with a new patch that follows Bellgrims suggestion.

larssilven:
Close this since #194 was created instead.

PR #193, opened 2016-03-01
EVP_aes_xxx_wrap functions must have been optionals

agrandville:


bellgrim:
#build

pspacek:
How is that possible that `HAVE_AES_KEY_WRAP_PAD` is supported but `HAVE_AES_KEY_WRAP` is not? How can I reproduce it?

PR #194, opened 2016-03-03
Add a slot with an uninitialized token after initializing an uninitia…

larssilven:


larssilven:
I will fix the test compilation errors. please wait.

bellgrim:
I made something similar to this, but in #198. Just to keep the slot handling within the SlotManager and to be more compliant with PKCS#11. Now the slot with an uninitialized token will only be added when calling C_GetSlotList with a NULL_PTR.

Would that work for you?

The calling application should:
* C_GetSlotList(NULL_PTR)
* C_GetSlotList(PTR)
* Find uninitialized token by using C_GetSlotInfo / C_GetTokenInfo
* C_InitToken
* C_GetSlotList(NULL_PTR)
* C_GetSlotList(PTR)
* Find uninitialized token by using C_GetSlotInfo / C_GetTokenInfo
* C_InitToken
* ...

larssilven:
Tested the #198 that also does what this is doing.

Issue #143, opened 2015-09-04
Delete a slot using softhsm2-util


bellgrim:
Nikos Mavrogiannopoulos:
While softhsm2-util --init-token --slot XX can be used to add more tokens in different slots in the soft token, there is no other option in softhsm2-util which allow to delete slots.

Migrated from:
https://issues.opendnssec.org/browse/SOFTHSM-117
https://issues.opendnssec.org/browse/SUPPORT-162

bellgrim:
In PKCS#11 there is no concept of deleting a token. So we need to a tool specific for SoftHSM.

All tokens are located in the directory given in the configuration file. The slot numbers are assigned according to the order of which the token directories are loaded from disc. The name of the token directory is a 16 byte random UUID.

When initializing a new token, its slot number will always be N+1. N is the number of existing tokens. But once the new token has been initialized and you reload the library, it is a high probability that it will be assigned a new slot number (because of the random directory name).

To avoid any confusion, e.g. people scripting `softhsm2-util --init-token --slot X` and `softhsm2-util --delete-token --slot X`, we should use the token label or token serial as the identifier.

We should also ask the user to verify the action (--force to ignore the question) and that the user login as the SO.

Or we could just document the manual procedure of removing the correct token. --show-slots will display the token serial which is the second half of the tokens directory name.

nmav:
I agree that slot numbers should be a bad reference for the actual tokens. Being able to delete with the token label or serial would be sufficient. It may also be good to remove the requirement to specify a slot number to initialization (i.e., softhsm2-util --init-token --slot 0), and simply initialize the next free token.

jwalzer:
Question: Is this somewhere stated in the documentation? 
- Are slot-sumbers stable?
- Should they be stable?
- Labels are not forced to be unique and I see problems incoming, if it would be forced 
- So how do I guarantee a stable identification to my Keys over the lifetime of my store?
- Will the serial-number be guaranteed to be stable?

Currently we're in the process of evaluating SoftHSM2 for DNSSEC with PowerDNS and what makes it quite difficult is the fact, that the slotnumber (which is also used to identify the slot) possibly can change when we add new slots.

jschlyter:
Keys, and other objects on a token, identified by a identifier (CKA_ID) and an optional label (CKA_LABEL). Tokens are identified by their label as well as their serial number. There is no real guarantee for uniqueness among any of these identifiers. However, tokens created by SoftHSM has certain properties.

[...]

PR #198, opened 2016-03-06
Add a slot with an uninitialized token by calling C_GetSlotList with NULL_PTR

bellgrim:
Calling C_GetSlotList with NULL_PTR will now make sure that there is always a slot with an uninitialized token.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -553,7 +553,7 @@ CK_RV SoftHSM::C_GetSlotList(CK_BBOOL tokenPresent, CK_SLOT_ID_PTR pSlotList, CK
 {
 	if (!isInitialised) return CKR_CRYPTOKI_NOT_INITIALIZED;
 
-	return slotManager->getSlotList(tokenPresent, pSlotList, pulCount);
+	return slotManager->getSlotList(objectStore, tokenPresent, pSlotList, pulCount);
 }
 
 // Return information about a slot
--- a/src/lib/slot_mgr/SlotManager.cpp
+++ b/src/lib/slot_mgr/SlotManager.cpp
@@ -64,24 +64,37 @@ SlotManager::~SlotManager()
 }
 
 // Get the slot list
-CK_RV SlotManager::getSlotList(CK_BBOOL tokenPresent, CK_SLOT_ID_PTR pSlotList, CK_ULONG_PTR pulCount)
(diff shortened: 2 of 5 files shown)


larssilven:
I think that you also should remove the creation of an empty slot after C_Initialize since it is not needed any more with you solution:
diff --git a/src/lib/slot_mgr/SlotManager.cpp b/src/lib/slot_mgr/SlotManager.cpp
index c63a759..8176a72 100644
--- a/src/lib/slot_mgr/SlotManager.cpp
+++ b/src/lib/slot_mgr/SlotManager.cpp
@@ -46,9 +46,6 @@ SlotManager::SlotManager(ObjectStore* objectStore)
                Slot* newSlot = new Slot(objectStore, i, objectStore->getToken(i));
                slots.push_back(newSlot);
        }
-
-       // Add an empty slot
-       slots.push_back(new Slot(objectStore, objectStore->getTokenCount()));
 }
 
 // Destructor

bellgrim:
You are not required to call C_GetSlotList before e.g. calling C_OpenSession. So it is good to have that initial slot added. One example is when working with command line utilities. The first command could be to display available slots (C_GetSlotList/C_GetSlotInfo/C_GetTokenInfo) then the next one could be used to interact with a given slot (C_OpenSession/C_Login/...). The empty uninitialized token is not persistent between the two different command calls, so we always need to make sure that there is one token available.

larssilven:
I see what you mean. Not a good idea to have an application that is depending on that there is an existing token on a certain slot ID. Slot IDs should always be fetched with the C_GetSlotList in the application before using them since p11 will not guarantee that you get same IDs when a new application is started. But there are applications that relies on that slot IDs are as before after application start and it is good if these apps also could use this p11. I'm actually looking on another change that will make the slot ID of an existing token the same even after a new token has been added.

So I'm totally satisfied with this. Thank you very much.

bellgrim:
Applications like OpenDNSSEC will use the token label to find the correct token. But simple tools like softhsm2-util use the slot number. There is a discussion in #143 to change this.

larssilven:
The sunpkcs11 wrapper for java used to access a p11 token as a java keystore is not capable of using the token label either. The token is referenced either with slot ID or index in slot list,

bellgrim:
#build

bellgrim:
#build again. Is the disc full on Ubuntu14?

bellgrim:
#build again. The are now more space available on ubuntu14.

PR #199, opened 2016-03-09
The slot ID of a token should always be the same

larssilven:
It would be good if a specific token always could be fetched from a slot with a specific ID.
Now the slot ID where a specific token is found is often changed after a new token is initialized.

The reason is that there are applications that are not using C_GetSlotList but are assuming that a token will always be in a slot with the same ID. Of course the best thing should be to change the application to use C_GetSlotList  and then find the slotID with serial or label. But unfortunate you don't always have the possibility to change the code. An example is the [SunPKCS11 java provider](http://docs.oracle.com/javase/8/docs/technotes/guides/security/p11guide.html#Config). Most java application that are using p11 tokens are using this provider.

This patch sets the slot ID to the 31 LSBs (the ID is configured as a java int that can not be negative in SunPKCS11) of the token serial. Uninitialized tokens will get slot with ID as before.

The patch is also containing #198 since I need to have it in the branch that I am testing.

### DIFF
--- a/src/lib/test/InfoTests.cpp
+++ b/src/lib/test/InfoTests.cpp
@@ -34,41 +34,9 @@
 #include <config.h>
 #include <stdlib.h>
 #include <string.h>
-#include <cppunit/extensions/HelperMacros.h>
 #include "InfoTests.h"
-#include "testconfig.h"
 
 CPPUNIT_TEST_SUITE_REGISTRATION(InfoTests);
-
-void InfoTests::setUp()
-{
-//    printf("\nInfoTests\n");
-
-#ifndef _WIN32
-    setenv("SOFTHSM2_CONF", "./softhsm2.conf", 1);
-#else
-    setenv("SOFTHSM2_CONF", ".\\softhsm2.conf", 1);
-#endif
-
-	CK_UTF8CHAR pin[] = SLOT_0_SO1_PIN;
-	CK_ULONG pinLength = sizeof(pin) - 1;
-	CK_UTF8CHAR label[32];
-	memset(label, ' ', 32);
-	memcpy(label, "token1", strlen("token1"));
-
-	// (Re)initialize the token
-	CK_RV rv = C_Initialize(NULL_PTR);
-	CPPUNIT_ASSERT(rv == CKR_OK);
-	rv = C_InitToken(SLOT_INIT_TOKEN, pin, pinLength, label);
-	CPPUNIT_ASSERT(rv == CKR_OK);
-	C_Finalize(NULL_PTR);
-}
-
-void InfoTests::tearDown()
-{
-	// Just make sure that we finalize any previous failed tests
-	C_Finalize(NULL_PTR);
-}
 
 void InfoTests::testGetInfo()
 {
@@ -164,19 +132,19 @@ void InfoTests::testGetSlotInfo()
 	// Just make sure that we finalize any previous failed tests
 	C_Finalize(NULL_PTR);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, &slotInfo);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_CRYPTOKI_NOT_INITIALIZED);
 
 	rv = C_Initialize(NULL_PTR);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, NULL_PTR);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, NULL_PTR);
 	CPPUNIT_ASSERT(rv == CKR_ARGUMENTS_BAD);
 
-	rv = C_GetSlotInfo(SLOT_INVALID, &slotInfo);
+	rv = C_GetSlotInfo(m_invalidSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_SLOT_ID_INVALID);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, &slotInfo);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_TOKEN_PRESENT ) == CKF_TOKEN_PRESENT);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_REMOVABLE_DEVICE ) == 0);
@@ -198,8 +166,6 @@ void InfoTests::testGetSlotInfoAlt()
     setenv("SOFTHSM2_CONF", ".\\softhsm2-alt.conf", 1);
 #endif
 
-	CK_UTF8CHAR pin[] = SLOT_0_SO1_PIN;
-	CK_ULONG pinLength = sizeof(pin) - 1;
 	CK_UTF8CHAR label[32];
 	memset(label, ' ', 32);
 	memcpy(label, "token1", strlen("token1"));
@@ -207,28 +173,33 @@ void InfoTests::testGetSlotInfoAlt()
 	// (Re)initialize the token
 	rv = C_Initialize(NULL_PTR);
 	CPPUNIT_ASSERT(rv == CKR_OK);
-	rv = C_InitToken(SLOT_INIT_TOKEN, pin, pinLength, label);
+	rv = C_InitToken(m_initializedTokenSlotID, m_soPin1, m_soPin1Length, label);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 	C_Finalize(NULL_PTR);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, &slotInfo);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_CRYPTOKI_NOT_INITIALIZED);
 
 	rv = C_Initialize(NULL_PTR);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, NULL_PTR);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, NULL_PTR);
 	CPPUNIT_ASSERT(rv == CKR_ARGUMENTS_BAD);
 
-	rv = C_GetSlotInfo(SLOT_INVALID, &slotInfo);
+	rv = C_GetSlotInfo(m_invalidSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_SLOT_ID_INVALID);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, &slotInfo);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_TOKEN_PRESENT ) == CKF_TOKEN_PRESENT);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_REMOVABLE_DEVICE ) == CKF_REMOVABLE_DEVICE);
 
 	C_Finalize(NULL_PTR);
+#ifndef _WIN32
+	setenv("SOFTHSM2_CONF", "./softhsm2.conf", 1);
+#else
+	setenv("SOFTHSM2_CONF", ".\\softhsm2.conf", 1);
+#endif
 }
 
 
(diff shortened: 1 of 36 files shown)


rijswijk:
I would like to comment on this; if I remember correctly, not returning
the same slot ID was a conscious decision, precisely because application
developers must not assume that the slot IDs are always the same. This
is clearly reflected in the PKCS #11 specification. Personally, I'm not
in favour of going down the road of working around bad application
developer behaviour, unless this is really blocking adoption of SoftHSM
v2. I leave it up to Rickard as main dev to make the final decision.

A note on the Java provider; their implementation seems to assume that
it is likely that there is only a single slot, in which case their
default behaviour of always selection the first slot in the array
returned by C_GetSlotList makes (twisted) sense. In general, I believe
SoftHSMv2 will return slots in the same order, so that could probably be
made to work with this implementation.

Lars Silvén wrote:
> It would be good if a specific token always could be fetched from a slot
> with a specific ID.
> Now the slot ID where a specific token is found is often changed after a
> new token is initialized.
> 
> The reason is that there are applications that are not using
> C_GetSlotList but are assuming that a token will always be in a slot
> with the same ID. Of course the best thing should be to change the
> application to use C_GetSlotList and then find the slotID with serial or
> label. But unfortunate you don't always have the possibility to change
> the code. An example is the SunPKCS11 java provider
> <http://docs.oracle.com/javase/8/docs/technotes/guides/security/p11guide.html#Config>.
> Most java application that are using p11 tokens are using this provider.
> 
> This patch sets the slot ID to the 31 LSBs (the ID is configured as a
> java int that can not be negative in SunPKCS11) of the token serial.
> Uninitialized tokens will get slot with ID as before.
> 
> The patch is also containing #198
> <https://github.com/opendnssec/SoftHSMv2/pull/198> since I need to have
> it in the branch that I am testing.
> 
> ------------------------------------------------------------------------
> 
> 
>         You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/opendnssec/SoftHSMv2/pull/199
> 
> 
>         Commit Summary
> 
>   * Calling C_GetSlotList with NULL_PTR will now make sure that there is
>     always a slot with an uninitialized token.
>   * now same slot ID as token serial when reading tokens after C_Initialize
> 
> 
>         File Changes
> 
>   * *M* src/lib/SoftHSM.cpp
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-0> (2)
>   * *M* src/lib/session_mgr/SessionManager.cpp
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-1> (8)
>   * *M* src/lib/session_mgr/SessionManager.h
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-2> (4)
>   * *M* src/lib/slot_mgr/Slot.cpp
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-3> (14)
>   * *M* src/lib/slot_mgr/Slot.h
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-4> (6)
>   * *M* src/lib/slot_mgr/SlotManager.cpp
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-5> (33)
>   * *M* src/lib/slot_mgr/SlotManager.h
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-6> (2)
>   * *M* src/lib/slot_mgr/test/SlotManagerTests.cpp
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-7> (137)
>   * *M* src/lib/slot_mgr/test/SlotManagerTests.h
>     <https://github.com/opendnssec/SoftHSMv2/pull/199/files#diff-8> (2)
> 
> 
>         Patch Links:
> 
>   * https://github.com/opendnssec/SoftHSMv2/pull/199.patch
>   * https://github.com/opendnssec/SoftHSMv2/pull/199.diff
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opendnssec/SoftHSMv2/pull/199>.
> 

-- 
-- SURFnet bv

larssilven:
Most HSMs and some smart cards offers more than one token so therefore most application (including java applications) ought to have support for several tokens.
The order is changed after a new token is added with C_Initialize or after a token backup from another HSM is restored or if a token is removed.
So if SoftHSM should be useable for java applications where tokens could be added or removed the slot ID has to be kept the same for a token.
Of course it is the java provider that should be changed, but making Oracle to change something in their java implementation takes years.

rijswijk:
Hi Lars,

Lars Silvén wrote:
> Most HSMs and some smart cards offers more than one token so therefore
> most application (including java applications) ought to have support for
> several tokens.
> The order is changed after a new token is added with C_Initialize or
> after a token backup from another HSM is restored or if a token is removed.
> So if SoftHSM should be useable for java applications where tokens could
> be added or removed the slot ID has to be kept the same for a token.

What if we keep the order the same but keep the behaviour w.r.t. the
slot ID undefined? That should work for this Java provider, right?

> Of course it is the java provider that should be changed, but making
> Oracle to change something in their java implementation takes years.

Yes, that's a typical case of "when hell freezes over" unfortunately.

@bellgrim: can you decide how you want to proceed?

Cheers,

Roland

-- 
-- SURFnet bv

larssilven:
The disadvantage with keeping the order will be that then you can not delete a token. Just add new ones that then has to be placed last.
I actually looked on this possibility for my patch but found it easier to just set the slot ID to a part of the serial than to track the initialization time of the token that you otherwise has to do.
What are the disadvantages of using a part of the serial number as slot ID except that it adds some complexity is not needed for apps that follows the standard.

Cheers,
Lars

rijswijk:
Hi Lars,

Lars Silvén wrote:
> The disadvantage with keeping the order will be that then you can not
> delete a token. Just add new ones that then has to be placed last.
> I actually looked on this possibility for my patch but found it easier
> to just set the slot ID to a part of the serial than to track the
> initialization time of the token that you otherwise has to do.
> What are the disadvantages of using a part of the serial number as slot
> ID except that it adds some complexity is not needed for apps that
> follows the standard.

My principal objection to it would be that it's a bit of a dirty hack
that (ab)uses one property to achieve another property. It would be
cleaner to store the slot ID with the token as a hidden attribute
somewhere in the object store and then to use that. This would, however,
require a global store somewhere that keeps track of a monotonically
increasing slot ID that is incremented each time a new token is initialised.

Cheers,

Roland

-- 
-- SURFnet bv

larssilven:
Hmmm,
I can not really not see that it abuses anything; you are free to set the slot ID to whatever you want as long as it is unique.
The probability that it should not be unique is very low even with just 31 bits if the serial number is random which I believe it is.
Having a slot ID that is a part of some unique constant part of the token will not break anything for any well written application just help some badly written applications.
You decide the way you want. But we would rather want to use the main SoftHSM instead of doing a fork of it that we have to keep in sync with the main.

Cheers,
Lars

rijswijk:
Hi Lars,

Lars Silvén wrote:
> I can not really not see that it abuses anything; you are free to set
> the slot ID to whatever you want as long as it is unique.
> The probability that it should not be unique is very low even with just
> 31 bits if the serial number is random which I believe it is.

So the token serial is a randomly generated GUID. The probability of
that being non-unique is 1/2^128, i.e. negligible. While it is likely
that the last 31 bits are also unique within a single SoftHSM instance,
that probability is not negligible, so you would need to include some
code to ensure its uniqueness. It is probably sufficient to assert(..)
on non-unique slot IDs, but you want the code to show defined behaviour
if this occurs.

> Having a slot ID that is a part of some unique constant part of the
> token will not break anything for any well written application just help
> some badly written applications.
> You decide the way you want. But we would rather want to use the main
> SoftHSM instead of doing a fork of it that we have to keep in sync with
> the main.

In principle I see no objection to adopting this behaviour, you have
made a case for it based on the (poorly written) Java provider. But the
final decision I leave to @bellgrim, my advice would now be to adopt
this behaviour, but implement a uniqueness check (assertion is fine, as
long as the failure mode is well-defined).

Cheers,

Roland

-- 
-- SURFnet bv

larssilven:
> So the token serial is a randomly generated GUID. The probability of that being non-unique is 1/2^128, i.e. negligible. While it is likely that the last 31 bits are also unique within a single SoftHSM instance, that probability is not negligible, so you would need to include some code to ensure its uniqueness. It is probably sufficient to assert(..) on non-unique slot IDs, but you want the code to show defined behaviour if this occurs.

I will do some assert even if I'm sure that we will all die long before there will be 2 different tokens anywhere that will share the same slot ID :-)

Cheers,
Lars

bellgrim:
What if we change the token UUID to contain a timestamp in the first number of bytes?

The slots will then be numbered in creation order. The newest token will get the highest slot number. However, if you remove a token then the tokens created after it will be renumbered.

larssilven:
Yes if you sort by creating time the slot index of trailing tokens will change if a token is removed. Also the order will likely change if you move or copy a token from one HSM to another.
So I prefer having a unique slot ID for each token

Cheers,
Lars

larssilven:
Another positive side effect of the assertion is that we now always will be sure that each token in the HSM really has an unique serial number.
For example; some bug could maybe create a new token with the same serial if you import a backup of a token when this token is already in the HSM.

bellgrim:
> Yes if you sort by creating time the slot index of trailing tokens will change if a token is removed. Also the order will likely change if you move or copy a token from one HSM to another.
So I prefer having a unique slot ID for each token

I was not thinking of the file property but to encode a time stamp in the beginning of the file name. Then new tokens will be added to the end of the list, given enough resolution of the time stamp.

E.g.
```
+// Generate a new token UUID string
+std::string UUID::newTokenUUID()
+{
+       RNG* rng = CryptoFactory::i()->getRNG();
+
+       ByteString uuid;
+
+       if (!rng->generateRandom(uuid, 12))
+       {
+               ERROR_MSG("Fatal, could not generate random UUID");
+
+               throw -1;
+       }
+
+       // Current time
+       time_t rawTime;
+       time(&rawTime);
+
+       // The ByteString is stored as a 64-bit value
+       ByteString byteTime(rawTime);
+
+       // Convert it to a string
+       char uuidStr[37];
+
+       if (byteTime[0] != 0 || byteTime[1] != 0 ||
+           byteTime[2] != 0 || byteTime[3] != 0)
+       {
+               sprintf(uuidStr, "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
+                       byteTime[0], byteTime[1], byteTime[2], byteTime[3],
+                       byteTime[4], byteTime[5],
+                       byteTime[6], byteTime[7],
+                       uuid[0], uuid[1],
+                       uuid[2], uuid[3], uuid[4], uuid[5], uuid[6], uuid[7]);
+       }
+       else
+       {
+               sprintf(uuidStr, "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
+                       byteTime[4], byteTime[5], byteTime[6], byteTime[7],
+                       uuid[0], uuid[1],
+                       uuid[2], uuid[3],
+                       uuid[4], uuid[5],
+                       uuid[6], uuid[7], uuid[8], uuid[9], uuid[10], uuid[11]);
+       }
+
+       return std::string(uuidStr);
+}
```

I believed that the tokens was sorted in alphabetical order, but the order returned by readdir is not guaranteed and is equal to the one returned by `ls -f`

So we also need to sort the directory vector.

```
diff --git a/src/lib/object_store/Directory.cpp b/src/lib/object_store/Directory
index 934ce94..df3fdb4 100644
--- a/src/lib/object_store/Directory.cpp
+++ b/src/lib/object_store/Directory.cpp
@@ -35,6 +35,7 @@
 #include "OSPathSep.h"
 #include "log.h"
 #include <string>
+#include <algorithm>
 #include <vector>
 #ifndef _WIN32
 #include <dirent.h>
@@ -215,6 +216,8 @@ bool Directory::refresh()
     finished:
 #endif
 
+       std::sort(subDirs.begin(), subDirs.end());
+
        valid = true;
 
        return true;
```

larssilven:
I understood you right from the beginning; some sort of sorting of the token dirs.
I will try to clarify the problem that I see with this:

- If any token that is not last will be removed then the index in the slot list for some of the remaining tokens are altered.

- If a token (files) created on a cloned HSM (using same token encryption keys) is installed it will alter the list if is not younger than all other tokens.

The second concern comes from that we want to have the possibility to create a cluster of HSMs. Each node in the cluster should normally have the same set of tokens. But the nodes will not be in sync all the time. When the nodes are out of sync they must be synced. In the syncing process tokens may be transfered between the nodes.
Apart from this issue I believe the SoftHSM suites the need in our project well and is much better than other available alternatives. So it should be very nice if there could find a good solution that makes everyone happy

bellgrim:
Ok, lets go with setting the slotID to the token serial.

This would also prioritize adding token label / serial support to the support tools. #143

larssilven:
Thank you Rickard,

I will now see to that the now failing test will pass.

Cheers,
Lars

larssilven:
But another test is failing. Seems to be a system test that is assuming that there always is an initialized token on a slot with ID 0.
I will use C_GetSlotList to fix this.

larssilven:
Now I'm satisfied with my changes.
Please review and tell me if there is anything that needs to be changed.
If you want I could create a new branch with only one commit for this and then do a new "pull request".

Cheers,
Lars

src/lib/test/InfoTests.cpp
```diff
@@ -207,28 +173,33 @@ void InfoTests::testGetSlotInfoAlt()
...
 	CPPUNIT_ASSERT(rv == CKR_SLOT_ID_INVALID);
 
-	rv = C_GetSlotInfo(SLOT_NO_INIT_TOKEN, &slotInfo);
+	rv = C_GetSlotInfo(m_notInitializedTokenSlotID, &slotInfo);
 	CPPUNIT_ASSERT(rv == CKR_OK);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_TOKEN_PRESENT ) == CKF_TOKEN_PRESENT);
 	CPPUNIT_ASSERT((slotInfo.flags & CKF_REMOVABLE_DEVICE ) == CKF_REMOVABLE_DEVICE);
 
 	C_Finalize(NULL_PTR);
+#ifndef _WIN32
```
bellgrim, line 198: Just guessing that this was a copy-paste error. Or do we need to set the environment variables here?

bellgrim:
#build

bellgrim:
The changes are looking good. Just my comment above.

src/lib/test/InfoTests.cpp
larssilven, line 198: No this is intentional.
Before it was a setUp() method in each test class. All these methods was setting the "SOFTHSM2_CONF" to the default "./softhsm2.conf". This was done before each test since setUp() is executed before each test.
Now there are no setUp() in any of the test classes. Instead each test class inherits the setUp() from any of 2 base classes (TestsBase or TestsNoPINInitBase). Instead of doing the setting in TestsNoPINInitBase::setUp() that is executed before each test the setting of the default is only done twice:
- in p11test.cpp before start of the testing.
- in this method after an alternative has been tested. There is a setting of an alternative to be tested in the beginning of this method. Here it is set back to the default.

larssilven:
Good that you liked it.
The change that you had a remark on was intentional. See my comment on your comment above.

src/lib/test/InfoTests.cpp
bellgrim, line 198: Ahh yes, missed that it was for setting the original configuration.

bellgrim:
#build

Issue #186, opened 2016-02-03
"Funny behaviour" when using C_InitToken() from GnuTLS' p11tool


vanrein:
I am not getting what I expected when using the GnuTLS generic utility `p11tool` for token initialisation (of an already existing token):

    GNUTLS_PIN=1234 p11tool --provider /path/to/softhsm2.so --initialize 'pkcs11:yada;yada;yada'

This calls `C_InitToken()` on the slot holding the token found in the slot referenced by `pkcs11:yada;yada;yada`:

 * https://gitlab.com/gnutls/gnutls/blob/gnutls_3_4_9/src/p11tool.c#L289
 * https://gitlab.com/gnutls/gnutls/blob/gnutls_3_4_9/src/pkcs11.c#L886
 * https://gitlab.com/gnutls/gnutls/blob/gnutls_3_4_9/lib/pkcs11_write.c#L1168
 * https://gitlab.com/gnutls/gnutls/blob/gnutls_3_4_9/lib/pkcs11_int.c#L244

The `C_InitToken()` function requires the SO-PIN, but after entering it `p11tool` consistently fails.  It seems that the token may be completely destroyed by the call, which is not what I would have expected.  To be honest, I am confused what is really going on.  My expectation was that the token would continue to exist, but with all private keys removed and the user PIN removed.

I am assuming that SoftHSMv2 is the problem, because tokens seem to be removed.  But I cannot be completely certain, of course.

bellgrim:
Running C_IninToken() on an already initialized token will re-initialize it. Thus removing all objects.

SoftHSMv2 does this by removing the directory of the token and initializing a new token in the same slot. Maybe we should just remove the objects and set the new token label?

vanrein:
Re-initialisation is the expected behaviour, but it does not seem to get out just the right way.  May caches are reset, or locks abolished?  I cannot make out if the token directory itself would be removed, or just its contents, but that is probably the difference you are talking about here.  Removing the token directory could easily cause such hick-ups of course.  Or producing a new random name for that directory.

I'm afraid I can't answer whether this would solve the behaviour I detected.  Is there anything I can do to test this further, and help to resolve the detected behaviour?

larssilven:
I checked the p11tool that happened to be on my laptop.
The token urls (your pkcs11:yada;yada;yada) look like this for my softhsm tokens:
[...]

PR #204, opened 2016-03-29
Issue #186: Reinitializing a token will now keep the token

bellgrim:
But all token objects are deleted, the user PIN is removed and the token label is updated.


bellgrim:
#build

PR #207, opened 2016-04-14
Fix segmentation faults in loadLibrary function

jariq:
This patch fixes segmentation faults in loadLibrary function on Windows when PKCS#11 library is not found.


bellgrim:
#build

bellgrim:
Thanks for fixing this!

PR #209, opened 2016-04-19
Possibility to test other p11 implementation with the cppunit test.

larssilven:
I have modified the cppunit test so that it also can test other p11 implementations.
If the test is built as before (make p11test) you will get the same test that you had before (static linking of the SoftHSMv2 p11).
But with this change it will also be possible to build the test so that any p11 implementation with a shared library could be tested.
I have also added 2 new options to the test (no arguments will test as before):
- Printing of test name before a test is executed. This is good if the p11 that you are testing is crashes when some test is executed.
- Possibility to run a specific test. This is good to have when debugging

The new build and execution options are documented in: ./src/lib/test/README

I believe that it will be good for the SoftHSM project if the test is used for testing other p11 implementations. If the test is used for testing other implementations enhancement of the test might be done by those using it to test those implementations.

### DIFF
--- a/src/lib/test/AsymEncryptDecryptTests.cpp
+++ b/src/lib/test/AsymEncryptDecryptTests.cpp
@@ -101,18 +101,18 @@ void AsymEncryptDecryptTests::rsaEncryptDecrypt(CK_MECHANISM_TYPE mechanismType,
 		mechanism.ulParameterLen = sizeof(oaepParams);
 	}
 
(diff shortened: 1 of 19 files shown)


bellgrim:
#build

bellgrim:
Thank you for the extended p11 tests!

larssilven:
And thank you for merging it so fast. I am very pleased to be able to use the main project and not my own fork.

Issue #203, opened 2016-03-21
Change command utils to use label or serial and not slot number


bellgrim:
Initializing an empty token will probably change the slot order, so the user should not use the slot number for other commands than for initialization.

Update the command utils to use the label or serial to find the token and its slot number.

bellgrim:
Fixed in #211

PR #211, opened 2016-05-01
Issue #203: Update the command utils to use the token label or serial…

bellgrim:
… to find the token and its slot number.


bellgrim:
#build

PR #212, opened 2016-05-26
Fixing of no padding cipher. 

larssilven:
If "no padding" ciphering was done with C_EncryptUpdate or C_DecryptUpdate then the call failed when the updated part was not a multiple of the block size.
This is wrong; these functions must handle any part length.
The requirement of multiple block size does only affect C_EncryptFinal and C_DecryptFinal; when any of these functions is called then the total length must be a multiple of the block size.
This commit fixes these issues.
Also the prediction for the needed output buffer length that is returned when a crypto function is called with NULL_PTR as input should now be more accurate.
To do a thorough test of all this, the "p11test" has been enhanced to use different part lengths to the update function.


bellgrim:
#build

bellgrim:
Thank you for fixing the encrypt/decrypt. The changes looks correct. It is just that FreeBSD does not like the ulong data type.

bellgrim:
#build

PR #222, opened 2016-08-02
C_GenerateKeyPair: mark public key as non private by default

nmav:


rijswijk:
@bellgrim someone else is also asking about this (I have just replied to them)

@nmav does this solve an interoperability problem with particular software? See also SoftHSMv2 #223 where there is more information on the choice of making CKA_PRIVATE true or false.

mouse07410:
> does this solve an interoperability problem with particular software?

I'm not @nmav but since he's the one who submitted this PR, I'd assume that it does.

nmav:
@rijswijk the issues that I see with the previous behavior are the following:
1. It not an expected behavior and we had to change two tools (pkcs11-tool and p11tool) to set any generated public keys as non-private by default.
2. in pkcs#11 (for keys that are not RSA) it is not possible to obtain the public key parameters from a private key object handle; obtaining them requires an application to open the public key object and read it. If the pubkey is private, the latter means that the user will be prompted for his PIN twice, only to read public parameters. An alternative is for an application to cache the user's PIN (something that is not possible when there is a pinpad), but it is a very bad work-around.

For these reasons I think it makes a good default having the public key object as public.

bellgrim:
#build

bellgrim:
#build again

mouse07410:
I don't mean to be a pest - but could you please share your plans for merging this PR?

Thanks!

bellgrim:
#build

As soon as the build environment has been fixed. The FreeBSD servers were down.

PR #232, opened 2016-09-08
Fix openssl lib ordering to support static linking

matthauck:


matthauck:
Would be cool to add a travis ci build configuration to build against openssl statically, but not quite sure how to do that...

bellgrim:
#build

bellgrim:
Yes it would be good to add some tests with static linking. This would also verify that this patch is working. Will check into this later.

Issue #190, opened 2016-02-23
Support for OpenSSL 1.1.0


bellgrim:
There are changes in OpenSSL 1.1.0 that requires updated code in SoftHSM. Build, test and update the code so that we support both 1.0.0 and 1.1.0.

bellgrim:
Have a look in #183 for examples of issues that need to be fixed.

bellgrim:
Fixed in #234

PR #234, opened 2016-09-10
Issue #190: Support for OpenSSL 1.1.0

bellgrim:


bellgrim:
#build

jariq:
@bellgrim I am testing this patch on Windows and would like to add some changes (mostly related to win32 build process). However I am not sure whether to open PR against `bellgrim:issue190` branch or wait until this gets merged into `opendnssec:develop`. Any suggestions?

bellgrim:
Thanks for patching the Windows build. Forgot to update the Windows configure script and build files.

If it works I could merge the changes into my branch.

jariq:
You have fixed all the issues that bothered me before I even managed to put a PR together :) However I am not sure about the build notes because I was using slightly different command to configure OpenSSL:

`perl Configure VC-WIN32 --prefix=C:\build\bin\openssl-1.1.0a-x86 --openssldir=C:\build\bin\openssl-1.1.0a-x86\ssl no-shared`

Individual options:
- `openssldir` - without this option I had to run command line in administrator mode because by default those files got installed into `Program Files` folder
- `no-shared` - this option actually enables static build

BTW  I am not sure whether `enable-static-engine` option is still valid in 1.1.0. If I remember correctly the build was successful regardless of its presence. I'll try to investigate this a little more.

bellgrim:
Thanks, update the build notes according to your suggestions.

Issue #215, opened 2016-07-05
softhsm2-util fails to run on Mac OS X


mouse07410:
Platform: Mac OS X 10.11.5, Xcode-7.3.1, Macports-installed OpenSSL-1.0.2h
SoftHSMv2: 2.1.0 (current GitHub master)

Build process (after cloning):
```
$ bash ./autogen.sh
. . . . .
$ ./configure --prefix=/opt/local --enable-64bit
. . . . . [normal configuration]
$ make all && make check
. . . . . [normal build, all checks passed]
$ 
```

All the checks passed.

However, when I try to run `softhsm2-util --init-token --slot 0 --label "Soft Token 1"`, I'm getting this error:

```
$ softhsm2-util --init-token --slot 0 --label "Soft Token 1"
ERROR: Could not initialize the library.
$ softhsm2-util --version
2.1.0
$ otool -L /opt/local/bin/softhsm2-util
/opt/local/bin/softhsm2-util:
	/opt/local/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/opt/local/lib/libsqlite3.0.dylib (compatibility version 9.0.0, current version 9.6.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
$
```

System log has these (strange) entries:
```
7/5/16 19:36:45.480 softhsm2-util[8385]: SimpleConfigLoader.cpp(84): Could not open the config file: /Users/uri/softhsm.conf

7/5/16 19:36:45.480 softhsm2-util[8385]: Configuration.cpp(158): Failed to load the SoftHSM configuration

7/5/16 19:37:49.099 softhsm2-util[8388]: ObjectStore.cpp(59): Failed to enumerate object store in /opt/local/var/lib/softhsm/tokens/

7/5/16 19:37:49.099 softhsm2-util[8388]: SoftHSM.cpp(476): Could not load the object store
7/5/16 19:37:56.022 softhsm2-util[8389]: ObjectStore.cpp(59): Failed to enumerate object store in /opt/local/var/lib/softhsm/tokens/
```

Help would be appreciated!

mouse07410:
It appears that `softhsm2-util` can only be used by privileged user. I wish documentation mentioned that fact.

[...]

PR #235, opened 2016-09-11
add the libtool install command for OSX

markwylde:
Had to do this on El Capitan 10.11.6


bellgrim:
Thanks, will test this when I try to resolve the issue in #215

Issue #227, opened 2016-08-16
Support alternate configuration loaders


matthauck:
I'm trying to write some unit tests in a library that is using softhsm and I am finding it very difficult to guarantee a clean configuration state in between tests using the environment variable configuration method (SOFTHSM2_CONF). This is difficult in particular on account of softhsm linking against the MSVC runtime dynamically (see #226). 

I'd like to expose a function in softhsm that supplies an alternate `ConfigLoader` that `C_Initialize` could use in place of `SimpleConfigLoader`. Does this sound like an idea that softhsm would be willing to take in upstream?

jariq:
I don't understand what exactly is difficult with the environment variable approach. Could you please provide more details?

jariq:
It is rather a blind shot but could you please check whether [this patch](https://github.com/jariq/SoftHSMv2/commit/52e7af281fda250c30ad31fe551adfe67fac248e) solves your problem?

matthauck:
Sorry -- missed the updates here. I was thinking something more like this: https://github.com/matthauck/SoftHSMv2/commit/46aca4378f7b4675a700c119b712abcd4ed78da0. Not sure if that kind of thing may be interesting. Not sure the right way to expose it as an interface... 

We ended up working around it by setting the SOFTHSM2_CONF environment variable from the build system instead of from inside the test executable.

jariq:
I understand suggested approach with custom configuration loader however I am more interested in finding out why environment variable approach did not suit you.  Could you please describe what issues you were having with the environment variable?

matthauck:
I was trying to write a test suite that would dynamically set `SOFTHSM2_CONF` environment variable before each test so that each test had a fresh softhsm configuration. That did not work because whatever the SOFTHSM2_CONF environment variable was at _test executable startup time_ was cached and not modifiable later on. I believe this is because the softhsm2.dll is compiled with `/MD` and links against the runtime dynamically, whereas our test executables are compiled with `/MT` and link the runtime statically. I believe this leads to softhsm having a different view into the environment. 

Workaround: Set SOFTHSM2_CONF once from the build system to a fixed location, and then before each test just rewrite the configuration to point to a new/fresh tokens directory.

Does that make sense?

jariq:
[...]

PR #236, opened 2016-09-12
Use GetEnvironmentVariable instead of getenv on Windows

jariq:
Values of all environment variables seems to be cached when Visual C++ redistributable libraries are loaded into the process. Any later call to `getenv()` function returns only cached values regardless of any subsequent environment modifications. More detailed analysis of this phenomenon has been previously published at [codeproject.com](http://www.codeproject.com/Articles/43029/When-what-you-set-is-not-what-you-get-SetEnvironme).

I have personally encountered this IMO strange behavior in [pkcs11-logger project](https://github.com/Pkcs11Interop/pkcs11-logger) and fixed it easily by reading the values of environment variables with `GetEnvironmentVariable()` function instead of `getenv()` function. Proposed patch does the same for `SOFTHSM2_CONF` environment variable.

I believe this patch also fixes issues discussed in  #227 where @matthauck wants to inject custom ConfigLoader implementation as a workaround.

### DIFF
--- a/src/lib/common/SimpleConfigLoader.cpp
+++ b/src/lib/common/SimpleConfigLoader.cpp
@@ -240,20 +240,58 @@ static char *get_user_path(void)
 	return NULL;
 }
 
+static char *get_env_var_path(void)
+{
+#ifdef _WIN32
+
+	LPSTR value = NULL;
+	DWORD size = 0;
+
+	size = GetEnvironmentVariableA("SOFTHSM2_CONF", value, size);
+	if (size == 0) {
+		return NULL;
+	}
+	
+	value = (LPSTR) malloc(size);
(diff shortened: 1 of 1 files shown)


bellgrim:
#build

bellgrim:
Thanks, the patch is looking good. Will merge once we get feedback from @matthauck

bellgrim:
#build

matthauck:
Undid my workaround and verified that this patch does indeed allow the softhsm2.dll to retrieve the environment variable where it previously could not.

bellgrim:
#build

The Ubuntu14 build machine had no space left. Should have been fixed.

bellgrim:
Thanks for testing this patch. Great that it worked.

Issue #239, opened 2016-09-20
Crash on module unload


dwmw2:
I tried using the SoftHSMv2 module with 'openssl rsautl'...

```
 SOFTHSM2_CONF=softhsm2.conf PKCS11_MODULE_PATH=/home/dwmw2/git/SoftHSMv2/src/lib/.libs/libsofthsm2.so \
  gdb --args openssl rsautl -engine pkcs11 -keyform engine \
      -inkey 'pkcs11:token=openconnect-test;id=%01;pin-value=1234' -sign -in foo -out bar 
```
It works, but then dies on exit:

```
engine "pkcs11" set.

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff547e990 in ?? ()
(gdb) bt
#0  0x00007ffff547e990 in ?? ()
#1  0x00007ffff6e0c312 in CRYPTO_THREADID_current (id=0x7fffffffd490) at cryptlib.c:493
#2  0x00007ffff6ec6468 in ERR_remove_thread_state (id=id@entry=0x0) at err.c:997
#3  0x0000000000419d42 in main (Argc=<optimized out>, Argv=<optimized out>) at openssl.c:437
```
Environment is Fedora 24 (OpenSSL 1.0.2), with updated libp11/engine_pkcs11 and SoftHSMv2 from latest git. Not that the latter seem to make any difference.

dwmw2:
Adding a `CRYPTO_set_id_callback(NULL)` to OSSLCryptoFactory::~OSSLCryptoFactory() seems to fix this, although I'm not sure any of that is really well-behaved given the the *application* may be using OpenSSL and its own threading model.

dwmw2:
It looks like `CRYPTO_set_id_callback()` has been deprecated since OpenSSL 1.0.0 and shouldn't be needed at all. Since 1.0.0 *(actually commit https://github.com/openssl/openssl/commit/4c3296960de32e5abfbb8f4703a2ce624d82669f in 2008)* OpenSSL has had its own fallbacks which include using the address of `errno`. Surely that's going to work on any platform with pthreads?
And since we are only calling `CRYPTO_set_id_callback()` when `HAVE_PTHREAD_H` is set, that makes it entirely redundant?

bellgrim:
Thanks, the patch has been merged.

PR #240, opened 2016-09-21
Issue #239: Crash on module unload with OpenSSL

dwmw2:
We use CRYPTO_set_id_callback() to set a callback, but we don't ever
remove it again on unload. So OpenSSL crashes the next time it needs a
thread-id.

CRYPTO_set_id_callback() has been deprecated since OpenSSL 1.0.0, the
oldest we support. And redundant too, since OpenSSL has fallbacks which
include the address of errno. Which is going to work on any platform
with pthreads... and we were only calling CRYPTO_set_id_callback() on
platforms with pthreads.

So just rip it out.

### DIFF
--- a/src/lib/crypto/OSSLCryptoFactory.cpp
+++ b/src/lib/crypto/OSSLCryptoFactory.cpp
@@ -55,9 +55,6 @@
 #include "OSSLGOST.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


bellgrim:
#build

bellgrim:
Thanks for fixing this. The changes are looking good.

Issue #237, opened 2016-09-13
Add windows continuous integration build


matthauck:
Verifying the fix for #236, I noticed that it did not build on windows without modifying the `util` project in visual studio to link against `crypt32.lib`. Supporting windows is important for us, and it will be hard to enforce that it continues to work on windows moving forward unless the build / tests are run there frequently. Is there any way to hook up a windows node to build on your jenkins instance?

matthauck:
fwiw, the application I'm working on builds softhsm on each build and wasn't too hard to setup. We do something more or less like this on windows:

```
cd win32
python Configure.py disable-debug verbose without-cppunit with-crypto-backend=openssl with-openssl=<path-to-openssl> enable-64bit
msbuild /m /p:Configuration=Release /p:Platform=x64 softhhsm2.sln
```

You would also need an openssl build of course. You'd also probably want to build with cppunit instead of without it, but even just making sure it compiles on windows for each PR would be a step forward I think.

bellgrim:
Yes, that would be really good.

@halderen There is some work going on with Jenkins. Do we have any plans on adding a Windows machine?

jariq:
BTW we currently working with @peponeska on an automated SoftHSM2 build with [AppVeyor](https://www.appveyor.com/) service. It is free for open source projects, works quite well and is used for example by OpenSC project. We plan to open PR once it is finished and tested.

bellgrim:
Thanks for creating the AppVeyor config and build scripts.

peponeska:
Thanks for merging it.

PR #241, opened 2016-09-21
added EXTRALIBS to utils project

peponeska:
Added missing library reference in util project which prevented successful building on windows. This issue was mentioned also in #237.


bellgrim:
Thanks for the patch.

PR #249, opened 2016-09-30
Allow config path to be specified in C_Initialize() call

dwmw2:
The `pReserved` field of `pInitArgs` can be used in a token-specific way, to indicate the config path. This provides an alternative to setting the `SOFTHSM2_CONF` environment variable, which may be non-trivial _(or indeed impossible, if you want to do it from a multi-threaded program before loading the SoftHSM token)_.

This is exactly how the NSS softokn module does it, and p11-kit supports the setting of a string in this field by means of the `x-init-reserved:` directive in its module file, thus:

```
module:libsofthsm2.so
x-init-reserved:SOFTHSM2_CONF=/home/dwmw2/git/openconnect/tests/softhsm2.conf
```

Obviously it's entirely optional; it overrides the existing environment variable and `DEFAULT_SOFTHSM2_CONF` if provided, but doesn't have to be present.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -362,8 +362,15 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
 		// Must be set to NULL_PTR in this version of PKCS#11
 		if (args->pReserved != NULL_PTR)
 		{
-			DEBUG_MSG("pReserved must be set to NULL_PTR");
-			return CKR_ARGUMENTS_BAD;
+			static const char *confstr = "SOFTHSM2_CONF=";
+			char *init_args = (char *)args->pReserved;
+
+			if (strncmp(init_args, confstr, strlen(confstr)) != 0 ||
+			    init_args[strlen(confstr)] == '\0') {
+				DEBUG_MSG("pReserved must be set to NULL_PTR or \"SOFTHSM2_CONF=...\"");
+				return CKR_ARGUMENTS_BAD;
+			}
+			SimpleConfigLoader::i()->setConfigPath(init_args + strlen(confstr));
 		}
 
 		// Can we spawn our own threads?
--- a/src/lib/common/SimpleConfigLoader.cpp
+++ b/src/lib/common/SimpleConfigLoader.cpp
@@ -56,6 +56,9 @@ std::unique_ptr<SimpleConfigLoader> SimpleConfigLoader::instance(nullptr);
 std::auto_ptr<SimpleConfigLoader> SimpleConfigLoader::instance(NULL);
 #endif
 
+static char *get_user_path(void);
+static char *get_env_var_path(void);
+
 // Return the one-and-only instance
 SimpleConfigLoader* SimpleConfigLoader::i()
 {
@@ -70,7 +73,11 @@ SimpleConfigLoader* SimpleConfigLoader::i()
 // Constructor
 SimpleConfigLoader::SimpleConfigLoader()
 {
-	configPath = getConfigPath();
+	configPath = get_env_var_path();
+	if (configPath == NULL)
+		configPath = get_user_path();
+	if (configPath == NULL)
+		configPath = strdup(DEFAULT_SOFTHSM2_CONF);
 }
 
 // Destructor
@@ -286,23 +293,6 @@ static char *get_env_var_path(void)
 
 #endif
 }
-
-char* SimpleConfigLoader::getConfigPath()
-{
-	char* configPath = get_env_var_path();
-	char *tpath;
-
-	if (configPath != NULL) {
-		return configPath;
-	} else {
-		tpath = get_user_path();
-		if (tpath != NULL) {
-			return tpath;
-		}
-		configPath = DEFAULT_SOFTHSM2_CONF;
-		return strdup(configPath);
-	}
-} 
 
 char* SimpleConfigLoader::trimString(char* text)
 {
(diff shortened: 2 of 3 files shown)


src/lib/SoftHSM.cpp
```diff
@@ -362,8 +362,13 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
 		// Must be set to NULL_PTR in this version of PKCS#11
 		if (args->pReserved != NULL_PTR)
 		{
-			DEBUG_MSG("pReserved must be set to NULL_PTR");
-			return CKR_ARGUMENTS_BAD;
+			char *init_args = (char *)args->pReserved;
+
+			if (strncmp(init_args, "SOFTHSM2_CONF=", strlen("SOFTHSM2_CONF")) != 0) {
```
bellgrim: Should be strlen("SOFTHSM2_CONF=") or we will miss to compare the equal sign.

src/lib/SoftHSM.cpp
```diff
@@ -362,8 +362,13 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
 		// Must be set to NULL_PTR in this version of PKCS#11
 		if (args->pReserved != NULL_PTR)
 		{
-			DEBUG_MSG("pReserved must be set to NULL_PTR");
-			return CKR_ARGUMENTS_BAD;
+			char *init_args = (char *)args->pReserved;
+
```
bellgrim, line 367: We should also check that the length of the incoming string is larger than the length of "SOFTHSM2_CONF=". Just so that we at least get a single character when calling setConfigPath().

jariq:
> This is exactly how the NSS softokn module does it...

NSS module is breaking the specification for years.

[PKCS#11 v2.20](https://github.com/Pkcs11Interop/PKCS11-SPECS/raw/master/v2.20/pkcs-11v2-20.pdf) chapter 11.4:
> For this version of Cryptoki, the value of `pReserved` thereby obtained must be `NULL_PTR`; if it’s not, then `C_Initialize` should return with the value `CKR_ARGUMENTS_BAD`.

[PKCS#11 v2.30  draft 4/6] (https://github.com/Pkcs11Interop/PKCS11-SPECS/raw/master/v2.30_DRAFT/pkcs-11v2-30b-d6.pdf) chapter 11.4:
> For this version of Cryptoki, the value of `pReserved` thereby obtained must be `NULL_PTR`; if it’s not, then `C_Initialize` should return with the value `CKR_ARGUMENTS_BAD`.

[PKCS#11 v2.40](https://github.com/Pkcs11Interop/PKCS11-SPECS/raw/master/v2.40/pkcs11-base-v2.40-os.pdf) chapter 5.4 :
> For this version of Cryptoki, the value of `pReserved` thereby obtained MUST be `NULL_PTR`; if it’s not, then `C_Initialize` should return with the value `CKR_ARGUMENTS_BAD`.

src/lib/common/SimpleConfigLoader.cpp
```diff
@@ -85,7 +104,6 @@ bool SimpleConfigLoader::loadConfiguration()
 		free(configPath);
```
bellgrim: Also remove this free().

bellgrim:
Good catch @jariq, with SoftHSM we aim to follow the standard. I will then vote against this patch.

dwmw2:
Fixed according to @bellgrim's comments; thanks.

Regarding what the spec says about `pReserved`... you don't get to deal with that `MUST` part, since you are the PKCS#11 provider, and not the caller of it. Crypto libraries which load PKCS#11 tokens (including NSS, GnuTLS and OpenSSL+ENGINE) are all already capable of setting `pReserved` because there is a long-standing convention which allows *(and even requires)* it for certain modules. Sure, they MUST not, but they're naughty. And they're only doing it for modules which they *know* accept it and do something useful with it. 

So the relevant part of the @jariq's citations, for SoftHSM, is the bit which says *" if it’s not, then `C_Initialize` should return with the value `CKR_ARGUMENTS_BAD`."*

Now I was interpreting that with the RFC2119 definition of 'SHOULD', which basically means "unless you really don't want to". :) 
I know they don't explicitly reference RFC2119, but there still is a very clear distinction between that 'should' and the shouty 'MUST' in the previous sentence.,

Pragmatically, it makes sense. It's a convention which has existed for years and isn't going away, and it's used in a token-specific way. It makes no sense to pass anything in `pReserved` when you're loading an arbitrary module, of course, but when you *know* you're loading the NSS softokn, or when you *know* you're loading SoftHSM, it's perfectly reasonable.

matthauck:
Just wanted to note that this is another approach to a similar pain I was experiencing when I opened #227. Using an environment variable to configure is painful. Not sure that `pReserved` is the right answer. I still think a separate softhsm-specific exported function would be better than trying to stuff it into the PKCS11 api somewhere. This would also provide a way for applications to detect whether they were using softhsm or not, i.e. `dlsym(handle, "SOFTHSM_set_conf_file") != NULL`

dwmw2:
A new entry point isn't necessarily `dlsym ()`. There are at least three different platform methods of loading libraries and referencing functions therein. Not to mention the fact the the application won't have access to that `handle` you mention because it'll be handled internally by the PKCS#11 loading glue  (like p11-kit, libp11, NSS's SECMOD, etc.)

OTOH both p11-kit and NSS have configuration files for loading modules that already let you say "load this PKCS#11 provider with *<this>* string in `pReserved`. And it is easily passed through whatever loading glue is used, like libp11 etc.

bellgrim:
Yes, the wording of the requirement makes it sound like it is possible to do this. That a generic initialization must use NULL_PTR, but a SoftHSM specific initialization could utilize this parameter.

PR #259, opened 2016-11-01
Set CKA_DECRYPT/CKA_ENCRYPT flags on key import to true.

### LINKED ISSUE #257: RSA key-import sets CKA_ENCRYPT and CKA_DECRYPT statically
When I import a key-pair using the `softhsm2-util` that I previously generated with openssl, I can  neither use the public key for encryption nor the private key for decryption. This is because the importer sets the corresponding attributes (`CKA_ENCRYPT`/`CKA_DECRYPT`)  statically to `CK_FALSE`. See: https://github.com/opendnssec/SoftHSMv2/blob/c4ef48dadc1ee81b50fdebf1d2a9517047352adf/src/bin/util/softhsm2-util-ossl.cpp#L241
Is there any specific reason for it or could this considered to be a bug?

mdomke:
When importing keys with the `softhsm2-util` the private and public key objects were attributed so that they couldn't be used for de-/encryption. This pull-request sets the attributes

* `CKA_WRAP`/`CKA_ENCRYPT` for the public key, and
* `CKA_UNWRAP`/`CKA_DECRYPT` for the private key

to true. This is supposed to fix #257

### DIFF
--- a/src/bin/util/softhsm2-util-botan.cpp
+++ b/src/bin/util/softhsm2-util-botan.cpp
@@ -236,8 +236,8 @@ int crypto_save_rsa
 		{ CKA_ID,               objID,        objIDLen },
 		{ CKA_TOKEN,            &ckToken,     sizeof(ckToken) },
 		{ CKA_VERIFY,           &ckTrue,      sizeof(ckTrue) },
-		{ CKA_ENCRYPT,          &ckFalse,     sizeof(ckFalse) },
-		{ CKA_WRAP,             &ckFalse,     sizeof(ckFalse) },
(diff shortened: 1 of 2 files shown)


bellgrim:
Thanks, could you also make the corresponding changes in softhsm2-util-botan.cpp?

mdomke:
@bellgrim Of course.

bellgrim:
#build

bellgrim:
The changes are looking good. Will merge it once we have released 2.2.0. Currently only merging changes that need to be fixed in the release candidate.

PR #261, opened 2016-11-02
Add support for libeaycompat lib for FIPS on windows

matthauck:


bellgrim:
The build script does not define WITH_FIPS. Should we have a config option that enable this and also make the choice between libeay32.lib and libeaycompat32.lib.

Is libeaycompat32.lib always built? If not, then the changes breaks the openssl 1.1.0 detection mechanism in the config script. The "if libeay32.lib elif libcrypto.lib" is changed into "if libeaycompat32.lib elif libcrypto.lib"

matthauck:
`libeaycompat32.lib` is built only when FIPS is enabled in the openssl build. I believe it is built to present the same API as `libeay32.lib` rather than needing to call `FIPS_` version of everything.

bellgrim:
Yes, but not all algorithms are available when using the FIPS version of OpenSSL. We thus need to set  WITH_FIPS so that we disable them in SoftHSM.

The selection of libeaycompat32.lib should be nested within the IF-statement for libeay32.lib. Now it is breaking the OpenSSL version detection (see the original if / elif / else statements)

Is there a corresponding FIPS library for OpenSSL 1.0.0?

bellgrim:
Something similar to:

```
        if os.path.exists(os.path.join(openssl_lib, "libeay32.lib")):
            if enable_fips:
                if os.path.exists(os.path.join(openssl_lib, "libeaycompat32.lib")):
                    openssl_lib_name = "libeaycompat32.lib"
                    openssl_lib_dll = "bin\\libeaycompat32.dll"
                else
                    print("can't find OpenSSL FIPS library", file=sys.stderr)
                    sys.exit(1)
            else
                openssl_lib_name = "libeay32.lib"
                openssl_lib_dll = "bin\\libeay32.dll"
        elif os.path.exists(os.path.join(openssl_lib, "libcrypto.lib")):
            openssl_lib_name = "libcrypto.lib"
            if platform == 32:
                openssl_lib_dll = "bin\\libcrypto-1_1.dll"
            else:
                openssl_lib_dll = "bin\\libcrypto-1_1-x64.dll"
        else:
            print("can't find OpenSSL library", file=sys.stderr)
            sys.exit(1)
```

matthauck:
hmm. good question. I haven't tried it out with openssl 1.1 yet. I'm still on 1.0.2.

matthauck:
Thinking about this more, `libeaycompat32.lib` actually does not map exactly to `WITH_FIPS`, which triggers including of `<openssl/fips.h>` in various places and calling the `FIPS_*` apis directly -- explicitly _not_ what libeaycompat32.lib is meant for.

win32/Configure.py
```diff
@@ -651,13 +653,16 @@ def doconfig():
         if os.path.exists(os.path.join(openssl_lib, "libeay32.lib")):
             openssl_lib_name = "libeay32.lib"
             openssl_lib_dll = "bin\\libeay32.dll"
+        if os.path.exists(os.path.join(openssl_lib, "libeaycompat32.lib")):
```
bellgrim: Ok, if it is for 1.0, then it should be changed from "if" to "elif". This is so that we have the version detection order:
- "libeay32.lib" => OpenSSL 1.1
- "libeaycompat32.lib" => OpenSSL 1.0 FIPS
- "libcrypto.lib" => OpenSSL 1.0

matthauck: Right, indeed. Updated to `elif`

Issue #265, opened 2016-11-04
Derive shared secret with ECC


Ovomaltine85:
Hi,

I stored a private and public ECC key to the hsm Slot. If I check the supported mechanisms with pkcs11-tool the mechanism for CKM_ECDH1 is supported. 
But if I use the pkcs11-tool with the --derive option then it is stated that the mechanism is not supported. How can the public and private ECC keys be used to derive a shared secret out of the hsm?

regards!

Ovomaltine85:
I also put this issue to the forum of OpenSC because I thought it could also be a pkcs11-tool issue.

This is the reply of my issue: https://github.com/OpenSC/OpenSC/issues/898

Maybe you can help me, why the operation throws an CKR_TEMPLATE_INCOMPLETE error?

Thank you

dengert:
As also stated in https://github.com/OpenSC/OpenSC/issues/898#issuecomment-258619835
it could an issue with the CKA_VALUE_LEN which is optional. 

6312 // CKA_VALUE_LEN must be specified
6313 if (keyLen == 0)
6314 {
6315 INFO_MSG("Missing CKA_VALUE_LEN in pTemplate");
6316 return CKR_TEMPLATE_INCOMPLETE;
6317 }

PKCS#11 v2.40 (pkcs110curr-v2.40-cs01.pdf) has:

2.3.9 Elliptic curve Diffie-Hellman key derivation
"This mechanism derives a secret value, and truncates the result according to the CKA_KEY_TYPE
attribute of the template and, if it has one and the key type supports it, the CKA_VALUE_LEN attribute of
the template. (The truncation removes bytes from the leading end of the secret value.) The mechanism
contributes the result as the CKA_VALUE attribute of the new key; other attributes required by the key
type must be specified in the template."

This implies CKA_VALUE_LEN is optional.

Ovomaltine85:
Hi!
I introduced manually the changes you proposed to SoftHSM.cpp and OSSLUtil.cpp and built the application again.
If I perform now the following command:
`pkcs11-tool --module /usr/local/lib/softhsm/libsofthsm2.so --slot 1 -l -p 1234 --id 1234 --derive --input-file pubkey.der -m ECDH1-DERIVE --output-file shared`
I get the following error:
`Using derive algorithm 0x00001050 ECDH1-DERIVE
error: PKCS11 function C_DeriveKey failed: rv = CKR_GENERAL_ERROR (0x5)
Aborting.`
[...]

PR #266, opened 2016-11-08
ECDH PKCS#11 compatability 

dengert:
In response to #265  two issues of compatibility were found while trying to use libsofthsm2.so with OpenSC pkcs11-tool. 

(1) SoftHSMv2 required the template to C_DeriveKey to provide a CKA_VALUE_LEN with the size of the secret key. Examples in PKCS\#11 do not requite this and PKCS\#11 V2.40 implies that the implementation can fill in some of this information. For example if  CKA_VALUE_LEN is not set or set to 0, the ECDH can set the key size to that returned from the ECDH_DERIVE. If an AES key is requested then a 16, 24 or 32 byte key could be be returned.  

(2) PKCS\#11 2.40 points out that in V2.20 the format of the EC Point of the peer public key was not well defined, and some implementations expected a DER format, while others expected a RAW format. V2.40 points out PKCS\#11 implementations must accept the RAW format, and can accept the DER format.   

These 2 commits show that EC key in SoftHSM and a NIST Demo PIV card with a EC key can derive the same CKK_GENERIC_SECRET using the others public key. the OpenSC pkcs11-tool can use libsofthsm2.so  for access to SoftHSM and  opensc-pkcs11.so to access the NIST demo card. 

The pkcs11-tool needs a mod when used with SoftHSM to request CKA_SENSITIVE=FALSE in order to read the CKK_GENERIC_SECRET. (A PR to OpenSC is also being made for this.) 

Please have a look at these changes.


bellgrim:
(1) I agree on these changes.

(2) SoftHSM is at PKCS#11 v2.20. And as you say it is not defined which format the pPublicData should be in. We copy the incoming data to an EC public key, which currently expects a DER-encoded ECPoint. The function that you are modifying is also used to read other attributes which must be in DER-format. The changes would now make SoftHSM to accept those attributes in RAW-format. I do not see any other values that use the RAW-format in v2.40. Perhaps we should make a RAW to DER-converter that is called before https://github.com/opendnssec/SoftHSMv2/blob/4dea95c1c8cbce4c4e7d19c9a978a7d45c9b7e5a/src/lib/SoftHSM.cpp#L10228 ?

bellgrim:
The internal RAW to DER-converter would also solve the issue without involving the crypto backend.

dengert:
With regards to a "RAW to DER-converter"...
Applications that call C_DeriveKey to do ECDH may also have chosen DER or RAW which works with the  preffered PKCS\#11 module.  OpenSC's pkcs11-tool and opensc-pkcs11.so chose RAW. The problem shows up when trying to use different applications with different modules. It is further complicated by p11-kit. 

For backward compatibility, PKCS\#11 modules need to accept both DER and RAW. In  https://github.com/opendnssec/SoftHSMv2/pull/266/commits/71cdd9ea61001649eb188df1ed0d8c6ff444c76a it tries to determine if it is DER or RAW. There is a chance it could get it wrong! They both start with 0x04. Using the fact that the both parties have to be using the same curve could be used to test the length.  With OpenSSL the size of the public key is determined by EC_POINT_point2oct based on 
field_length = (EC_GROUP_get_degree(ecgroup) +7)/8 and if it is COMPRESSED or UNCOMPRESSED. 
Knowing what the length should be can be used to determine if the input is DER which is 2 or more bytes longer.
Something similar could be done in SoftHSM. I could do that or someone from SoftHSM could write it. 

I am going to fix OpenSC to accept either RAW or DER as well. For compatibility with current modules, I have a mod to OpenSC pkcs11-tool to let the user specify --derive-pass-der to have pkcs11-tool pass in a DER version of the public key. But this is just one application. I expect to test it with unmodified SoftHSM today.

bellgrim:
We should absolutely support both RAW and DER for pPublicData in C_DeriveKey(). But this should not be done in OSSL::byteString2pt() because that will make SoftHSM to accept the RAW-format for all public EC points. E.g. when calling C_CreateObject()

I can have a look at making this internal "RAW to DER"-converter to detect what format the data have and make the adjustments before sending it to crypto backend. Will have some time a few weeks.

bellgrim:
Have fixed this in #288. 

When it comes to CKA_SENSITIVE, PKCS#11 says that e.g. CKA_VALUE from CKK_GENERIC_SECRET or CKK_AES cannot be revealed if object has its CKA_SENSITIVE attribute set to CK_TRUE or its CKA_EXTRACTABLE attribute set to CK_FALSE.

PR #272, opened 2016-11-17
Fix c++11 check to look for unique_ptr

matthauck:
Strangely enough Mac 10.8 passes the other C++11 checks
but does not have std::unique_ptr.


bellgrim:
#build

bellgrim:
Thanks for fixing this.

There is a build failure on ubuntu14, but the error is just the regular "No space left on device" which happened after the build tests. Will merge the patch.

PR #279, opened 2017-01-04
Editorial changes from Mountain Lion to Sierra.

mikeneumann:
Updated command line tools package name and to current as of 4-Jan-2016.

Made note of brew Caveat for libtool prepending a ‘g’.

Removed AM_INIT_AUTOMAKE update to configure.ac as this is already done.

By trial and error, I found that both softhsm2.conf AND
softhsm2-alt.conf must be edited as shown in order to avoid various
assertion errors generated by p11test in InfoTest.cpp.

There are still some compile-time warnings that may/may not be an
issue. Most concerning would appear to be:

DB.cpp:77:17: warning: 'va_start' has undefined behavior with reference
types [-Wvarargs]
        va_start(args, format);
                       ^
All tests performed by ‘check’, however, pass.

### DIFF
--- a/OSX-NOTES.md
+++ b/OSX-NOTES.md
@@ -1,9 +1,9 @@
-# Building SoftHSMv2 on Mac OS X 10.8.5 (Mountain Lion)
+# Building SoftHSMv2 on macOS 10.12.3 (Sierra)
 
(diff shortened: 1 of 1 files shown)


bellgrim:
Thank you for the update!

PR #284, opened 2017-01-27
Test minor fixups

alonbl:
Hi,

Minor fixups for tests.
1. A missing log object that is required to resolve all symbols.
2. Switch to autoconf based cppunit detection.

The cppunit detection is performed using pkg-config standard detection, I see you have some proprietary methods, I suggest to sync these as well. If for some reason you prefer the proprietary methods I can do that, just explain why you are using those.

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -117,6 +117,9 @@ ACX_VISIBILITY
 # If we should install the p11-kit module
 ACX_P11KIT
 
+# cppunit setetings
+ACX_CPPUNIT
+
 # Set full directory paths
 full_sysconfdir=`eval eval eval eval eval echo "${sysconfdir}" | sed "s#NONE#${prefix}#" | sed "s#NONE#${ac_default_prefix}#"`
 full_localstatedir=`eval eval eval eval eval echo "${localstatedir}" | sed "s#NONE#${prefix}#" | sed "s#NONE#${ac_default_prefix}#"`
--- a/m4/acx_cppunit.m4
+++ b/m4/acx_cppunit.m4
@@ -1,0 +1,4 @@
+AC_DEFUN([ACX_CPPUNIT],[
+	PKG_PROG_PKG_CONFIG
+	PKG_CHECK_MODULES([CPPUNIT], [cppunit], [have_cppunit=yes], [have_cppunit=no])
+])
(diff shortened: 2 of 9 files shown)


bellgrim:
#build

alonbl:
#build

bellgrim:
PKG_PROG_PKG_CONFIG is not available on all plattforms

alonbl:
@bellgrim: you already commit into your source many m4 macros[1]. why can't we add pkg.m4 as well?

[1]https://github.com/opendnssec/SoftHSMv2/tree/develop/m4

alonbl:
@bellgrim: I can help with transition if you accept, first commit to transit to PKG_ macro, second this change on top.

bellgrim:
Found the m4 here: https://gnunet.org/svn/Extractor/m4/pkg.m4

But it is using a GPL license, which is not compatible with BSD used by SoftHSM.

alonbl:
I do not fully understand... there is no violation of gpl as you distribute the source with original copyright statement and do not modify it. As Gentoo maintainer I never seen an issue with that,  including cairo, firefox, dbus, freetype, harfbuzz I stopped here in my greps... and these are only packages that are installed on my system.

OK, I will remove the usage, however, please re-validate this at your end, it will simplify your package and will enable best practice configuration/usage.

bellgrim:
Ok, I am no expert in this area. Just that we want to be able to distribute SoftHSM under a BSD license.

bellgrim:
But they do write:

```
# As a special exception to the GNU General Public License, if you
# distribute this file as part of a program that contains a
# configuration script generated by Autoconf, you may include it under
# the same distribution terms that you use for the rest of that program.
```

So it should be ok to include it.

alonbl:
good! I just finished to remove it... what do you want to do? I can make the other libraries like openssl use pkg-config as well instead of the current proprietary and fixed detection.

bellgrim:
Another reason to why PKG_PROG_PKG_CONFIG is missing is because pkg-config is not available / installed on the selected system. So even if we add the m4, the binary pkg-config would still be missing. Right?

bellgrim:
We are currently migrating over to a new build environment. So we will see how it works there.

alonbl:
#293 merged.

PR #285, opened 2017-01-28
crypto: openssl: libressl support

alonbl:
Hi,

Support libressl with trivial changes, passes all tests.

BTW: Running programs during autoconf is highly discouraged as cross-compile will fail. Please consider to use only the pre-processor in order to detect algorithms and versions.

BTW2: If you run something while allowing to specify --with-xxx=prefix, you should also set LD_PRELOAD.

Thanks!

### DIFF
--- a/m4/acx_openssl_gost.m4
+++ b/m4/acx_openssl_gost.m4
@@ -22,7 +22,7 @@ AC_DEFUN([ACX_OPENSSL_GOST],[
 				OpenSSL_add_all_algorithms();
 
 				/* Load engines */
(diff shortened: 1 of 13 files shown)


bellgrim:
Thanks for adding support for libressl. Will have a look on this within a few weeks.

bellgrim:
#build

alonbl:
#build

alonbl:
@bellgrim: this should be trivial and it did pass the current CI (those that were available anyway).

alonbl:
ping? I would like to avoid patching this every time in gentoo.
Thanks!

PR #289, opened 2017-02-16
Minor fix of 'EVP_CipherFinal_ex' call

viktorTarasov:
EVP_CipherFinal_ex is called without shift for output pointer -- so that
the final padded block is written at the begining of the output buffer,
and not appended to the result of the previous EVP_CipherUpdate.

It works now as expected because "padding is handled by cipher mode separately ..."
and final padded block has zero length.


bellgrim:
#build

Thanks for spotting this.

bellgrim:
#build

PR #291, opened 2017-02-21
Change to enable builds and reports on new Jenkinks environment

halderen:
- more temporary files that are not git ignored
  - ROOT is a directory used on the new jenkins environment
- Add support for building on Slackware Linux distribution
- Botan version no longer available at that URL in building script
- let cpp unit tests output for XML file to be picked up be Jenkins
  environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,14 @@ test-driver
 *.log
 *.trs
 
+# Temporary files
+*~
+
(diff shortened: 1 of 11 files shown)


halderen:
@bellgrim for your incorporation or comments.

bellgrim:
Looking good, thanks!

Issue #297, opened 2017-03-21
Errors when building SoftHSMv2 on OSX - TestFactoryRegistry.h missing


kombatminipig:
I've been following the included installation guide for OSX, but always run into the following error when running ``make check``: 

```bash
Making check in src
Making check in lib
Making check in common
make[3]: Nothing to be done for `check'.
Making check in crypto
Making check in test
/Applications/Xcode.app/Contents/Developer/usr/bin/make  cryptotest
depbase=`echo cryptotest.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
	g++ -DHAVE_CONFIG_H -I. -I../../../..  -I./.. -I./../.. -I./../../common -I./../../cryptoki_compat -I./../../data_mgr -I./../../object_store -I./../../session_mgr -I./../../slot_mgr -I/usr/local/opt/openssl/include   -g -O2 -Wall -Wextra -std=c++11 -fvisibility=hidden -MT cryptotest.o -MD -MP -MF $depbase.Tpo -c -o cryptotest.o cryptotest.cpp &&\
	mv -f $depbase.Tpo $depbase.Po
cryptotest.cpp:33:10: fatal error: 'cppunit/extensions/TestFactoryRegistry.h' file not found
#include <cppunit/extensions/TestFactoryRegistry.h>
         ^
1 error generated.
make[5]: *** [cryptotest.o] Error 1
make[4]: *** [check-am] Error 2
make[3]: *** [check-recursive] Error 1
make[2]: *** [check-recursive] Error 1
make[1]: *** [check-recursive] Error 1
make: *** [check-recursive] Error 1
```

Does anybody know what I've missed?

mouse07410:
@LudovicRousseau I'm afraid this fix does not work. The missing file belongs to the `cppunit` package that has to be installed. In my case it's installed in `/opt/local`. Now SoftHSMv2 does not include the directories provided in the `CFLAGS`, and so fails to locate this include file:
```
. . . . .
Making check in session_mgr
Making check in test
/Applications/Xcode.app/Contents/Developer/usr/bin/make  sessionmgrtest
depbase=`echo sessionmgrtest.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
[...]

PR #301, opened 2017-04-04
Issue #297: Fix build with cppunit

LudovicRousseau:
@CPPUNIT_CFLAGS@ was not correctly used.
So the cppunit header files were not found if not installed in the
default folder.


bellgrim:
There is also a similar issue on FreeBSD 10.3

PR #302, opened 2017-04-04
Export PKCS#11 symbols from the library

LudovicRousseau:
On macOS (at least 10.12) the library symbols are not visible so the
library can't be used.

The definition of PKCS_API could be smarter.
I inspired myself from https://github.com/LudovicRousseau/PCSC/blob/master/src/misc.h#L47

### DIFF
--- a/src/lib/main.cpp
+++ b/src/lib/main.cpp
@@ -38,6 +38,12 @@
 #include "cryptoki.h"
 #include "SoftHSM.h"
 
+#if defined(__GNUC__) && \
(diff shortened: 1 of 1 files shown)


mouse07410:
I confirm that this PR fixes the #303 problem on MacOS and urge @bellgrim to merge it ASAP. Without it SoftHSMv2 doesn't just exhibit issues on Mac - it's completely broken. So please enable us on Mac to use the current code by merging this PR.

bellgrim:
Thanks for fixing this!

PR #330, opened 2017-06-26
Fix warnings on syslog

nmav:


nmav:
A previous attempt to fix that in 2.2.0 was by making the default /var/lib/softhsm/tokens/ having the sticky bit and being world-writable. Unfortunately using the sticky bit has quite some risks, and I find it much better practice, a module to be silent by default.

nmav:
That relates to bug reported against Fedora:
https://bugzilla.redhat.com/show_bug.cgi?id=1425135

bellgrim:
Thank you for the patch!

Issue #140, opened 2015-09-04
Support CKA_ALLOWED_MECHANISMS for key objects


bellgrim:
The CKA_ALLOWED_MECHANISMS is not supported for key objects. If it is added, then correct checks in the different crypto operations must be implemented.

Migrated from:
https://issues.opendnssec.org/browse/SOFTHSM-81

PR #336, opened 2017-07-30
Support for CKA_ALLOWED_MECHANISMS

bdhess:
Took a stab at closing out #140, since I'm looking for this support for a project I'm working on.  

There are a couple of corners that I cut, and welcome feedback about how I can improve:
* I didn't create a new table in sqlite to hold mechanisms; rather, overloaded the byte string table. It's not entirely clear to me why different object types need different tables. In any case, I'd be happy to do something different, but would need to understand what an appropriate migration path would look like for users who are using a released version.
* I did not add support for mechanism lists in wrap/unwrap templates.  That probably ought to be there for this feature to be considered complete.

I will openly confess that my C++ is pretty novice, so I'd suggest that you probably want to be quite thorough on review.  ~~Also, I know that there are some tab/spacing issues-- I'll fix those up on the next push.~~ (edit: fixed)

### DIFF
--- a/src/bin/dump/common.h
+++ b/src/bin/dump/common.h
@@ -39,6 +39,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <map>
+#include <set>
 #include <string>
 #include <stdexcept>
 #include <vector>
@@ -436,6 +437,7 @@ public:
 	uint8_t boolValue;
 	I ulongValue;
 	std::vector<uint8_t> bytestrValue;
+	std::set<I> mechSetValue;
 
 	// Dump an array (in fact an Attribute vector) value
 	void dumpType() const;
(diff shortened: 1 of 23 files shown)


bellgrim:
Thank you for this pull request, I will review it within a couple of weeks.

bellgrim:
It looks good, also added the changes to softhsm2-dump-file. softhsm2-dump-db did not need any changes, since the data is stored as byte string table.

We currently only support basic attributes, and not array attributes, in the wrap templates. So we could save that for later.

bdhess:
Awesome, thanks for the review @bellgrim!  Is there anything you're looking for from me in order to get this merged?

I don't need to specify mechanism lists in wrap/unwrap templates, so at least given my n of 1 sample, there's value to this as-is. :)

bellgrim:
I also added support for CKA_ALLOWED_MECHANISMS in array attributes. This feature should now be considered complete.

PR #341, opened 2017-09-20
softhsm2-util is extended with --aes flag

cherezov:
In addition to existing ability to import key pair to soft hsm new flag --aes is added to indicate that imported file must be used as key value content and treated as AES secret key wile importing to HSM.


bellgrim:
Thank you for the pull request. I made some minor changes: include CKA_ID when creating the object and update the documentation.

Issue #141, opened 2015-09-04
Support CKA_ALWAYS_AUTHENTICATE for private key objects


bellgrim:
The attribute CKA_ALWAYS_AUTHENTICATE is accepted by private key objects, but no action is made when this attribute is set to true.

Migrated from:
https://issues.opendnssec.org/browse/SOFTHSM-82

bellgrim:
Fixed in #364

PR #364, opened 2017-12-22
Issue #141: Support CKA_ALWAYS_AUTHENTICATE for private key objects.

bellgrim:


src/lib/test/ObjectTests.cpp
```diff
@@ -1975,12 +2103,12 @@ void ObjectTests::testAllowedMechanisms()
 
 	CK_KEY_TYPE keyType = CKK_GENERIC_SECRET;
 	CK_OBJECT_CLASS secretClass = CKO_SECRET_KEY;
-	CK_BYTE key[64];
+	CK_BYTE key[65] = { "0000000000000000000000000000000000000000000000000000000000000000" };
```
Jakuje, line 2106: This looks like some separate change unrelated to the actual issue. Can you put it into separate commit?

jariq:
I did not review the code but I have tested following scenarios:

- Signature creation with `CKM_RSA_PKCS` and `CKM_RSA_PKCS_PSS` mechanisms
  - using RSA key that has `CKA_ALWAYS_AUTENTICATE` attribute set to `CK_FALSE`
  - using RSA key that has `CKA_ALWAYS_AUTENTICATE` attribute set to `CK_TRUE`  
- Signature creation with `CKM_ECDSA` mechanism
  - using EC key that has `CKA_ALWAYS_AUTENTICATE` attribute set to `CK_FALSE`
  - using EC key that has `CKA_ALWAYS_AUTENTICATE` attribute set to `CK_TRUE`

Everything worked as expected. Good job 👍

src/lib/test/ObjectTests.cpp
bellgrim: This was to fix a Valgrind memory error so that there were no issues when running the ObjectTests. Have split this into two commits.

PR #376, opened 2018-03-02
Add CKM_GENERIC_SECRET_KEY_GEN Support

### LINKED ISSUE #375: Support for CKM_GENERIC_SECRET_KEY_GEN
Currently using HMAC with PKCS11 2.4.0 is difficult as SoftHSM does not support generating generate secret generation via CKM_GENERIC_SECRET_KEY_GEN that many software packages require for using HMAC. Is there a planned addition for this in the near future?

Thanks.

daviddob:
Add support for CKM_GENERIC_SECRET_KEY_GEN which is required for HMAC operations in PKCS11 2.4.0.

Resolves #375

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -5639,107 +5639,7 @@ CK_RV SoftHSM::C_GenerateKey(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMecha
 
     if (pMechanism->mechanism == CKM_GENERIC_SECRET_KEY_GEN)
     {
-        const CK_ULONG maxAttribs = 32;
-        CK_ATTRIBUTE keyAttribs[maxAttribs] = {
-            { CKA_CLASS, &objClass, sizeof(objClass) },
-            { CKA_TOKEN, &isOnToken, sizeof(isOnToken) },
-            { CKA_PRIVATE, &isPrivate, sizeof(isPrivate) },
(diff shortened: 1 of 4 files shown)


gbolo:
nice, would love to have `CKM_GENERIC_SECRET_KEY_GEN` support in softhsm

gbolo:
Just tested this, I can confirm that it works as expected. Good job

bellgrim:
Thank you for this PR. Will check it out once I have finished some other issues in the backlog.

npmccallum:
Any update on this? We (Red Hat) have started adding PKCS#11 support to  https://github.com/latchset/jose but we are running into this bug.

One comment about the implementation. I'm not sure it actually conforms to http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cs01/pkcs11-curr-v2.40-cs01.html#_Toc399398911 . My cursory glance over the code looks like it is supplying attributes besides `CKA_KEY_TYPE` and `CKA_CLASS`. https://github.com/opendnssec/SoftHSMv2/pull/376/commits/0393f7a97ef1a9c90cb735d05e383efff15600de#diff-ff3900333e6090272c9e40bb1fc2da27R6900

I don't think that is correct. Also, I don't see validation of the attributes `CKA_KEY_TYPE` and `CKA_CLASS` when specified in the template.

bellgrim:
Made some changes so that we use the correct key type and also handle the KCV-attribute.

Issue #366, opened 2018-01-16
configure: cannot check for file existence when cross compiling


the-nic:
When I run configure to cross-compile SoftHSM 2.3.0, I get the following error message:

```
configure:5347: checking for crypto backend
configure:6032: result: Botan
configure:6052: checking for /usr/include/botan-2/botan/version.h
configure:6058: error: cannot check for file existence when cross compiling
```

Command line for configure:
`./configure --build=x86_64-linux --host=arm-poky-linux-gnueabi --target=arm-poky-linux-gnueabi --prefix=/usr --exec_prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --libexecdir=/usr/lib/softhsm --datadir=/usr/share --sysconfdir=/etc --sharedstatedir=/com --localstatedir=/var --libdir=/usr/lib --includedir=/usr/include --oldincludedir=/usr/include --infodir=/usr/share/info --mandir=/usr/share/man --disable-silent-rules --disable-dependency-tracking --with-libtool-sysroot=SOMEDIR/imx7dsabresd --with-crypto-backend=botan --with-botan=/usr`

Any ideas?

the-nic:
This error is likely caused by `AC_CHECK_FILE()` in `m4/acx_botan.m4`.

Maybe a `test -f <file>` would be enough, as suggested [in a mailing list](https://lists.gnu.org/archive/html/autoconf/2000-10/msg00018.html)?

bellgrim:
Thanks for reporting this. The suggested fix sounds good.

the-nic:
Note that in current HEAD the other `acx_botan_*.m4` files have a similar issue. Instead of using `AC_RUN_IFELSE` one could do `AC_COMPILE_IFELSE` and convert the test program to something like this:

```
int main()
{
    using namespace Botan;
#if BOTAN_VERSION_CODE < BOTAN_VERSION_CODE_FOR(2,0,0)
    #error "Unsupported feature"
#endif
    eturn 0;
}
```

Probably even only the preprocessor would be sufficient, but I am by no means an expert for autotools.

rijswijk:
If this is easy to fix, do it in 2.5.0

bellgrim:
Fixed with #381 and #410

PR #381, opened 2018-03-25
Ease cross-compilation

michaelweiser:
These patches make cross-compilation possible, easier and more robust. Changes are:

- do not by default add build host paths to lib and include search paths
- make runtime checks guess their outcome and overrideable
- switch some runtime checks to compile checks
- split the 64bit check to be more robust
- prefer pkg-config because it can be and on at least some Linux distros (Gentoo, Debian) is already wrapped for cross-compilation

This should make some headway with https://github.com/opendnssec/SoftHSMv2/issues/366.


bellgrim:
Great work, will check it once I have finished some other issues in the backlog.

PR #387, opened 2018-04-12
Add VS2017 detection to Configure.py

jariq:
Hi, I've successfully [built SoftHSM 2.4.0 with Visual Studio 2017 Community](https://github.com/disig/SoftHSM2-for-Windows/blob/2.4.0/BUILDING.md) but I had to pass `with-toolset=v141` parameter to `Configure.py` script. This micropatch adds VS2017 detection to `Configure.py` script.


bellgrim:
Thanks for adding this!

PR #388, opened 2018-04-13
Update DBTests.cpp to fix x86 test failure

tcely:
I don't know why this wasn't a problem on `x86_64` builds, but this test failed on `x86` alpinelinux.

See: https://github.com/tcely/aports/blob/a6fdb70c45eb7ce9eff1dff0993eeb93a012bfb6/testing/softhsm/fix-objstoretest.patch
http://build.alpinelinux.org/buildlogs/build-edge-x86/testing/softhsm/softhsm-2.4.0-r0.log


bellgrim:
Thank you for fixing this!

PR #395, opened 2018-05-03
Add SHA-256 hashing kdf for ECDH derivation

christopherhex:
Added the possibility to use a SHA-256 kdf when deriving ECDH keys


bellgrim:
Hi, sorry for a late review but I have been on a parental leave for the last 3 months.

Implementing CKD_SHA256_KDF requires us to derive according to ANSI X9.63. This pull request will only perform hashing of the derived secret, which is not sufficient. We also need to handle pSharedData, create a loop and hash the derived secret with a counter and the optional pSharedData, and output a concated version of the different hashed that matches the request key length.

See section 3.6.1 in http://www.secg.org/sec1-v2.pdf

Issue #401, opened 2018-06-21
Windows Setup: No key available for 4EE17CD2


chenx832:
When I do the following command, there's no key to receive. Could you please upload a public key?

win32Setup: line 219
gpg --keyserver pgp.mit.edu --recv-keys 4EE17CD2

Thanks!

bellgrim:
@halderen Could you have a look at this?

halderen:
The mentioned key and keyserver are actually incorrect.  The correct keyserver is pgp.surfnet.nl and the key-id is Issue #401, so the command mentioned in the documentation should actually have been:

    gpg --keyserver pgp.surfnet.nl --recv-keys 4FCB0B94

PR #402, opened 2018-06-25
Issue #401 wrong key and keyserver mentioned in installation documentation

halderen:
Issue #401
Out key isn't located at pgp.mit.edu but on the keyserver pgp.surfnet.nl.
At the moment it isn't easy to upload to all keyservers.
But also they key id is actually an old key, the new key in use for some
time is 4FCB0B94.


bellgrim:
Thank you for fixing this!

PR #407, opened 2018-07-18
Correct the default config path

dmjones:
According to `man softhsm2.conf` and experimentation on my system, the default config path is `/etc/softhsm/softhsm2.conf`.


dmjones:
Apologies, this was invalid. Perhaps my original installation was from an older copy of your source.

PR #423, opened 2018-10-19
crypto: use ENGINE_load_rdrand with recent openssl

alonbl:
libressl and probably older openssl do not support this.


bellgrim:
Thanks for the PR!

PR #434, opened 2018-12-11
CMake: fix build issues

Lekensteyn:
*    CMake: Fix shared library dependencies
    
    The autotools build links libsofthsm.so to OpenSSL, but the CMake one
    does not do this. This can be observed with `ldd` and GnuTLS 3.6.5 on
    Arch Linux:
    
        $ LD_DEBUG=all p11tool --provider /.../softhsm/libsofthsm2.so
        ...
         25469:     /.../softhsm/libsofthsm2.so: error: symbol lookup error: undefined symbol: DSA_set0_key (fatal)
        ...
        pkcs11_add_provider: PKCS #11 initialization error.
    
    Fix this by adding OpenSSL as dependency for the softhsm_crypto object
    library and subsequently propagate this to the softhsm library. Remove
    unused static library files while at it.
    
    This requires CMake 3.12, older versions do not allow targets to be
    linked with object libraries via target_link_libraries. (If this version
    bump is not desired, a workaround needs to be created for
    softhsm_objectstore and softhsm_crypto).

*    CMake: do not install static libsofthsm2 with ENABLE_STATIC=OFF
    
    Previously static and shared libraries were unconditionally built and
    installed. This removes the unused ENABLE_SHARED option and makes the
    ENABLE_STATIC functional by not installing libsofthsm2.a.
    (softhsm2-static still needs to be built for use by softhsm2-util, etc.)
    
    (--disable-static is a standard autotools option, but CMake does not
    have this as the project has to be explicit about building static or
    shared libraries.)

*    CMake: restore support for CMake 3.5
    
    Tested with CMake 3.5.2 (built from source).


Lekensteyn:
I am sorry for the confusion due to bad commit messages. I have rewritten them, hopefully they should be clearer now (the actual patches remain unmodified).

> you just pulled out support for Shared Libraries - libsofthsm2.so

This is a misunderstanding, I only drop the ENABLE_SHARED option because it was not checked anyway. So shared libraries are always built (as before).

> CMake 3.12.vs 3.5, there isn't anything in these CMake files that will actually prevent you from working with even CMake 2.8 other than having to put version checks and if-else blocks around some sections of the code to prevent warnings in newer versions.

This patch uses a feature from CMake 3.12 ([release notes](https://cmake.org/cmake/help/v3.12/release/3.12.html)) which was implemented via [Enable OBJECT library "linking"](https://gitlab.kitware.com/cmake/cmake/merge_requests/1524). From the release notes:
> The target_link_libraries() command now supports Object Libraries. Linking to an object library uses its object files in direct dependents and also propagates usage requirements.

If you the version bump is not acceptable for you, let me know and I'll adjust it and not rely on the feature (it'll result in uglier code though).

I hope that the updated commit message clarifies my environment and answers your question about how to reproduce it.

conz27:
@Lekensteyn , okay I understand what you did there ... was a bit confused earlier.

I'd say maintaining backwards compatibility is important just because not all supported environments have a trivial path of upgrading to newer version of CMake. I propose having both solutions implemented with a CMake version switch.
```
if(CMAKE_VERSION VERSION_LESS "3.12")
... # < 3.12 impl
else(CMAKE_VERSION VERSION_LESS "3.12")
... # 3.12+ impl
endif(CMAKE_VERSION VERSION_LESS "3.12")
```

I know it's not going to be the cleanest looking code now, but one day when CMake 3.12+ becomes standard everywhere, ripping out the old code will a trivial exercise.

src/lib/CMakeLists.txt
```diff
@@ -32,40 +32,33 @@ set(SOURCES access.cpp
...
+        set_target_properties(${PROJECT_NAME}-static
+                              PROPERTIES OUTPUT_NAME ${PROJECT_NAME}
+                              )
+        generate_export_header(${PROJECT_NAME}-static)
+endif()
 
 ###############################################################################
 # Shared Library Config
 ###############################################################################
-add_library(${PROJECT_NAME} SHARED ${SOURCES} ${OBJECT_FILES})
```
conz27, line 67: To fix linker problem, why not add:
```
target_link_libraries(${PROJECT_NAME} ${OPENSSL_LIBRARIES})
```

conz27: Nevermind, I like your suggestion better. 

Place the dependency on libsofthsm_crypto and then propagate up.

Lekensteyn:
I had to restore building softhsm-static since it was used by other targets (like softhsm-util). Also restored CMake 3.5 support. It's not very nice, but it seems to work.

src/lib/crypto/CMakeLists.txt
```diff
@@ -135,7 +135,9 @@ endif(WITH_BOTAN)
 include_directories(${INCLUDE_DIRS})
 
 add_library(${PROJECT_NAME} OBJECT ${SOURCES})
-target_link_libraries(${PROJECT_NAME} ${CRYPTO_LIBS})
+if(NOT CMAKE_VERSION VERSION_LESS "3.12")
```
conz27, line 140: How about:
```
if(CMAKE_VERSION VERSION_GREATER "3.11")
```
The above is more confusing to look at. If accepted, please change all instances.

Lekensteyn: I would use `VERSION_GREATER_EQUAL` if possible, but that unfortunately that requires at least https://cmake.org/cmake/help/v3.7/release/3.7.html, hence this (semantically equivalent) workaround.

`CMAKE_VERSION VERSION_GREATER "3.11"` would not be sufficient as it also matches 3.11.4. I could use `CMAKE_VERSION VERSION_GREATER "3.11.99"` if you think that is easier to understand?

conz27: You're right, they have widely distributed point releases from time to time (i.e. CMake 3.0.2) ... 

I wish `VERSION_GREATER_EQUAL` existed much earlier.

Your original change is probably the least of all evils given my workaround had a flaw - leave it as is.

Maybe just add a reference to the comment in `src/lib/CMakeLists.txt` to explain the existence of the if-else block, because it looks very strange without the relevant context. (apply to all affected files)

bellgrim:
Thank you for the pull request. Will now merge it.

PR #438, opened 2019-01-04
build: detect crypto algorithms by default

alonbl:
Currently the crypto algorithm settings are either enabled or disabled, this
makes it difficult to support permutations openssl, libressl, botan versions,
as these differ in capabilities.

In most cases, the library should use the crypto library capabilities unless
explicitly required otherwise by the users.

This change handles tree state value for the --enable-<algo>:
 * detect [default] - set to yes/no based on supported detection.
 * yes - must be supported and used.
 * no - should not be used.


bellgrim:
Thank you for the PR. Will merge it.

Issue #442, opened 2019-01-25
OpenSSL resources from OSSLEVPSymmetricAlgorithm should be released on plugin close


pemensik:
I did play a bit with BIND's dnssec-keyfromlabel tool with a combination of OpenSSL. I found it has problem with too late destruction of some unrelated object.

```
dnssec-keyfromlabel -E pkcs11 -a RSASHA256 -l 'pkcs11:token=DNS;object=DNS2;pin-value=1234' test
```

ended with backtrace 
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff72ff895 in __GI_abort () at abort.c:79
#2  0x00007ffff7ce6165 in isc_assertion_failed (file=file@entry=0x7ffff7f50a20 "../../../lib/dns/openssl_link.c", 
    line=line@entry=166, type=type@entry=isc_assertiontype_insist, 
    cond=cond@entry=0x7ffff7f50a00 "dst__memory_pool != ((void *)0)") at ../../../lib/isc/assertions.c:52
#3  0x00007ffff7efebe9 in mem_free (ptr=<optimized out>, file=<optimized out>, line=<optimized out>)
    at ../../../lib/dns/openssl_link.c:165
#4  mem_free (ptr=<optimized out>, file=<optimized out>, line=<optimized out>) at ../../../lib/dns/openssl_link.c:165
#5  0x00007ffff702979c in OSSLEVPSymmetricAlgorithm::~OSSLEVPSymmetricAlgorithm (this=0x5555555f5af0, __in_chrg=<optimized out>)
    at OSSLEVPSymmetricAlgorithm.cpp:52
#6  0x00007ffff701fe53 in OSSLAES::~OSSLAES (this=0x5555555f5af0, __in_chrg=<optimized out>) at OSSLAES.h:45
#7  OSSLAES::~OSSLAES (this=0x5555555f5af0, __in_chrg=<optimized out>) at OSSLAES.h:45
#8  0x00007ffff70324ac in SecureDataManager::~SecureDataManager (this=0x5555555c2600, __in_chrg=<optimized out>)
    at SecureDataManager.cpp:96
#9  0x00007ffff7032679 in SecureDataManager::~SecureDataManager (this=0x5555555c2600, __in_chrg=<optimized out>)
    at SecureDataManager.cpp:93
[...]

PR #443, opened 2019-01-25
442 open ssl free early

pemensik:
Proposal to fix issue #442 

Fixes a crash when used from BIND tools.


rijswijk:
Hi Petr,

It seems the CI checks fail, looks like the P11Test fails with these changes. I will ask @bellgrim to have a look if he has time.

PR #445, opened 2019-02-08
Add wrap support with CKM_AES_CBC

martinpaljak:


rijswijk:
Hi Martin, thank you for your contribution. We have a couple of requests before we can merge the pull requests:

- Can you also ensure that unwrapping is supported? It seems that the current code is only for wrapping. This is because we would prefer to keep the mechanism support complete.

- Can you also write testcases to check that the mechanism works correctly? You can simply extend the SymmetricAlgorithmTests in src/lib/test

Thanks!

martinpaljak:
Will look into it when time permits.

martinpaljak:
FYI: I've generalized this to add unwrap support "the proper way" (not in softhsm2.cpp but only for openssl aes driver) and extended to support the same for 3des. Not sure if I ever get to writing tests, as I test via Java.

Also, added not-so-nice-looking support for CKM_[XOR|CONCATENATE]_* that should probably be added as a separate PR...

Issue #449, opened 2019-03-04
Explicitly set CKA_ID to zero-length byte string on PRIVATE_KEY causes assertion on retrieval


space88man:
I am getting the following assertion on testing corner cases where `CKA_ID` is explicity set to 
zero-length byte string. It works for public keys and certs, but not for private keys.

```
/usr/include/c++/8/bits/stl_vector.h:950: std::vector<_Tp, _Alloc>::const_reference std::vector<_Tp, 
_Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) const [with _Tp = unsigned char; _Alloc = 
SecureAllocator<unsigned char>; std::vector<_Tp, _Alloc>::const_reference = const unsigned char&; 
std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)
```
SoftHSM2v2 is 2.5.0 (from Fedora 29 x86_64)

How to reproduce;

1. Create a private key on token (don't use  `CKA_ID` in the template). Observe that retrieving `CKA_ID` works: returns zero-length byte string.
2.  `C_SetAttributeValue` to set `CKA_ID` explicitly to zero-length byte string
3.   Retrieve `CKA_ID` to a non-NULL buffer of length 1 (this is what `pkcs11-tool` would do)
4. Get the above assertion

The first call (with NULL_PTR buffer works)
```
9: C_GetAttributeValue
2019-03-04 13:26:07.348
[in] hSession = 0x1
[in] hObject = 0x2
[in] pTemplate[1]: 
    CKA_ID                0000000000000000 / 0
[out] pTemplate[1]: 
    CKA_ID                0000000000000000 / 0
Returned:  0 CKR_OK
```

When you make  the second call pointing to a non-NULL buffer, the assertion appears.

space88man:
Backtrace:
```
#0  0x00007ffff7dff53f in raise () from /lib64/libc.so.6
#1  0x00007ffff7de9895 in abort () from /lib64/libc.so.6
#2  0x00007ffff7a42638 in std::__replacement_assert (
    __file=__file@entry=0x7ffff7a5fcb8 "/usr/include/c++/8/bits/stl_vector.h", __line=__line@entry=950, 
[...]

PR #450, opened 2019-03-05
Perform copy only if value has size > 0.

space88man:
Addresses #449.

When `value` has size 0, the invalid reference  `&byteString[0]` in `ByteString::const_byte_str` will cause an assertion when built with the flag `-Wp,-D_GLIBCXX_ASSERTIONS` (used by default in distros like Fedora)

### DIFF
--- a/src/lib/P11Attributes.cpp
+++ b/src/lib/P11Attributes.cpp
@@ -340,8 +340,10 @@ CK_RV P11Attribute::retrieve(Token *token, bool isPrivate, CK_VOID_PTR pValue, C
 					ERROR_MSG("Internal error: failed to decrypt private attribute value");
 					return CKR_GENERAL_ERROR;
 				}
-				const unsigned char* attrPtr = value.const_byte_str();
-				memcpy(pValue,attrPtr,attrSize);
+				if (value.size() !=  0) {
+					const unsigned char* attrPtr = value.const_byte_str();
+					memcpy(pValue,attrPtr,attrSize);
+				}
 			}
 			else if (attr.getByteStringValue().size() != 0)
 			{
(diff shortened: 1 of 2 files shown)


space88man:
Did you build with `-Wp,-D_GLIBCXX_ASSERTIONS`  on Fedora?

halderen:
No, I did not, but was busy doing so.  With that flag, the error occurs not just on Fedora, but on most (all?) platforms.  I think it might be wise to add the flag when building on our test platforms, so we can catch these errors.
Thanks for digging in and getting this flag.

space88man:
Fedora also uses `-Wp,-DFORTIFY_SOURCE=2`

https://access.redhat.com/blogs/766093/posts/1976213

space88man:
What about solving the issue in `ByteString.cpp`?

```If byteString.size() != 0 return &byteString[0] else return NULL```

Alternatively , if we don't want to ever get a NULL back (that might break existing users) some sentinel value in the translation unit

```
// if size() == 0 then return a pointer to this sentinel value
unsigned char sentinel[1]; //somewhere inside ByteString.cpp
```

space88man:
Some discussion from @halderen, @rijswijk  lost due to a forced push, recording here:

---------------------
@halderen: "I was more looking at a similar patch using
return byteString.data();
Which is protected against a zero sized string."

---------------------
@rijswijk: "if std::vector.data() can be cast to either unsigned char* or const unsigned char* that is probably the best solution (we may have originally used that and changed it to &byteString[0] later because that cast cannot be made on every platform, but it's too long ago to remember. The STL description says that in case of size() == 0 then .data() may or may not return a NULL pointer."

---------------------
@rijswijk: "Scrap that, we didn't try the .data() approach, since that didn't appear until the C++ 11 standard according to the STL docs. This could be a consideration, it may break builds on pre- C++ 11 systems."

halderen:
I think we can use the .data() call as discussed earlier, since C++11 has become the norm for most.  However I will pull in the issue as-is for now, perhaps indeed changing it into .data() later in order to make sure this fix will get into the next release in one way or another.

PR #455, opened 2019-03-07
Provide a new configuration option to allow enabling and disabling various mechanisms

Jakuje:
This is helpful for testing applications using PKCS#11 devices,
which do not have to provide all the variety of the mechanisms
provided by SoftHSM and might need to implement some checks
and possible fallback when detecting the algorithm to be used in the
upper protocol layers.

This is an addition to the `CKA_ALLOWED_MECHANISMS` attribute,
which is unfortunately not very widely used. It should serve the same
purpose better, but most of the tokens and PKCS#11 modules do not
provide it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ src/lib/session_mgr/test/sessionmgrtest
 src/lib/slot_mgr/test/slotmgrtest
 src/lib/test/p11test
 src/lib/test/softhsm2-alt.conf
+src/lib/test/softhsm2-mech.conf
 src/lib/test/softhsm2.conf
 src/lib/test/tokens/64d6c3fe-1575-1736-1d26-5ccb28440ea7/
 src/lib/test/tokens/dummy
--- a/configure.ac
+++ b/configure.ac
@@ -215,6 +215,7 @@ AC_CONFIG_FILES([
 	src/lib/test/Makefile
 	src/lib/test/softhsm2.conf
 	src/lib/test/softhsm2-alt.conf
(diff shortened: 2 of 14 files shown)


halderen:
Thank for your feature additions.  Looks useful, and code looks good.  I'm looking what causes the build on windows to fail, but I have very little info on how this works.
The one thing I would like to change though, is the actual softhsm2.conf parameter name.  I'll try to think of an alternative, because I do not want to use "token." here.  It is good that it is comma separated.  Will follow up to this.

Jakuje:
Thank you for the interest. The fail looks like coming from the test, at least in one random task:
```
1) test: InfoTests::testGetMechanismListConfig (F) line: 343 ..\..\src\lib\test\InfoTests.cpp
assertion failed
- Expression: rv == CKR_OK
```
It looks like I missed a windows change that copies the correct file into the place in `win32/p11test/p11test.vcxproj.in`. I will push updated changes soon.

Jakuje:
Ping. Is there anything more needed before merging of this pull request?

rijswijk:
Ah no, there isn't, we are planning to look at merging a bunch of PRs later this week.

halderen:
I can't make changes to the pull request to rename the token.mechanisms configuration parameter to a different name (which I need to do for future changes).  Hence I will create a different pull request for this.  I will rename the configuration parameter "token.mechanisms" to "slots.mechanisms".  My rationale is that the configuration parameter overrides whatever mechanisms are reported by the token itself, which is reported by the "slot" object in SoftHSM.

PR #462, opened 2019-03-27
Fix user PIN flags manipulation in DBToken

Aearsis:
The parentheses in the logic expression were wrong, not clearing the flags which should be cleared when user PIN is set.

In SoftHSMv2, this has the only effect of clearing USER_PIN_COUNT_LOW when user PIN is changed/initialized, as the other flags are never set. In OSToken, these flags are manipulated correctly.


halderen:
This was indeed a bug and the fix is proper.  Minor editing for readability which also would have make the bug more obvious.  Will merge, thanks.

Issue #464, opened 2019-04-03
Race condition with multiple threads on the same slot


Takarth:
When you close a session the session handler checks if no more sessions are open. If this is the case it releases the handle lock only to lock in immediately again and then releases all remaining handles to the corresponding slot. 

  

Scenario:  

- Thread 1 wants to close his session while Thread 2 wants to open a new session.  

- Both sessions use the same slot id. 

  

Thread 1 acquires handles lock, closes his session and sees there are no more session on the slot. It then releases the lock only to retake it again in the function allSessionsClosed(). But Thread 2 is faster and acquires the handles lock in the short period where it is not locked by Thread 1 and adds his newly created session. Thread 1 then acquires the lock again and destroys the session of thread 2. Resulting in CKR_INVALID_SESSION_HANDLE in subsequent calls for session 2. 

  

Functions: 

void HandleManager::sessionClosed(const CK_SESSION_HANDLE hSession) @ HandleManager.cpp 

void HandleManager::tokenLoggedOut(const CK_SLOT_ID slotID) @ HandleManager.cpp

rijswijk:
Do you perhaps have a test program that reproduces this so we can test any fix we implement?

Takarth:
Test code

#include "cryptoki.h"
#include <string.h>
#include <pthread.h>
#include <string>

void run() {
    CK_RV rv = CKR_OK;
    while(rv == CKR_OK) {
        CK_SESSION_HANDLE hSession = CKR_SESSION_HANDLE_INVALID;
        rv = C_OpenSession(0, CKF_RW_SESSION, nullptr, 0, &hSession);
        if(rv) {
            printf("C_OpenSession failed %d\n", rv);
            rv = CKR_OK;
            continue;
        }
        rv = C_CloseSession(hSession);
        if(rv) printf("C_CloseSession failed %d\n", rv); // Abort thread
    }
}

static void* thread1(void* context) {
   run();
   return nullptr;
}

static void* thread2(void* context) {
   run();
   return nullptr;
}

int main(int argc, char *argv[]) {
   std::string tokenLabel("test"), so_pin("hello");

   CK_C_INITIALIZE_ARGS ckArgs;
   memset(&ckArgs, 0x0, sizeof(CK_C_INITIALIZE_ARGS));
   ckArgs.flags = CKF_OS_LOCKING_OK;

   CK_RV rv = C_Initialize(&ckArgs);
[...]

PR #465, opened 2019-04-03
Do not release handlesMutex when no more open session exist on the slot (Issue #464)

Takarth:
Possible solution to my reported issue 464


rijswijk:
It looks like this PR does indeed solve the issue, I will ask @halderen to review it and approve it if he agrees.

Issue #469, opened 2019-04-15
C_EncryptFinal - segmentation fault


johughes99:
In testing out various error conditions in the C_Encrypt* set of function calls I noticed I get a segmentation fault when i pass a NULL ptr for the pulLastEncryptedPartLen parameter. Looking at the code in SoftHSM.cpp I notice there is no check for it being NULL

I also get some CKR_OK returns for some situations where I'm passing in either NULL pointers or zero length to the various function calls. But given the strategy in SoftHSMv2 seems to be towards leniency I'm not reporting those:-)

rijswijk:
I have implemented a fix and a test case which my colleague will review and merge probably later this weeks. Thanks for the report!

PR #470, opened 2019-04-16
Implemented fix and testcase for missing NULL_PTR validation

rijswijk:
This PR implements a fix for issue #469 and adds a test case to check if ```C_EncryptFinal``` and ```C_DecryptFinal``` correctly return ```CKR_ARGUMENTS_BAD``` when a ```NULL_PTR``` is provided in the return length field.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -2694,6 +2694,9 @@ CK_RV SoftHSM::C_EncryptFinal(CK_SESSION_HANDLE hSession, CK_BYTE_PTR pEncrypted
 {
 	if (!isInitialised) return CKR_CRYPTOKI_NOT_INITIALIZED;
 
+	// Github issue #469, check NULL_PTR on pulEncryptedDataLen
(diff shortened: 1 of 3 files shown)


halderen:
Indeed checking the parameters for NULL would be better here, and BAD_ARGUMENTS would be the proper return code.  The PKCS#11 specification does not indicate whether this argument may be null or not.  There is in fact a possible use case, and that is explicitly passing bad arguments to abort the multi-part operation.  In that case BAD_ARGUMENTS is a good return code anyway.
But in any case the multi-part operation must be terminated, as per specification of PKCS#11: "A call to C_DecryptFinal always terminates the active decryption operation unless it returnsC KR_BUFFER_TOO_SMALL or is a successful call (i.e., one which returns CKR_OK) to determine the length of the buffer needed to hold the plaintext."

This needs to be included, so I cannot pull in this change at the moment, as the current implementation does not terminate the operation.

rijswijk:
Committed the requested change.

halderen:
The unit test incorrectly tests if the operation can be continued after an BAD_ARGUMENTS, but the operation is aborted at this time.  Will fix this in subsequent PR.

PR #475, opened 2019-04-30
Rename configuration option to "slots.mechanisms".

halderen:
My rationale is that the configuration parameter overrides whatever
mechanisms are reported by the token itself, which is reported by the
"slot" object in SoftHSM.


Jakuje:
I am sorry I did not follow up on this comment in the original PR. I am fine with this new naming. Thank you for taking care of this.

PR #479, opened 2019-05-22
Increase sqlite busy timeout from 15 seconds to 3 minutes

jluebbe:
When many parallel accesses happen on a network-backed storage (like a
busy Ceph RBD), the DB can be busy for longer than 15 seconds, leading
to failed token accesses. Increase the timeout to 3 minutes (a factor of
12).

### DIFF
--- a/src/lib/object_store/DB.cpp
+++ b/src/lib/object_store/DB.cpp
@@ -848,7 +848,7 @@ bool DB::Connection::connect(const char *
 		return false;
 	}
 
-	rv = sqlite3_busy_timeout(_db, 15000); // 15 seconds
+	rv = sqlite3_busy_timeout(_db, 180000); // 3 minutes
 	if (rv != SQLITE_OK) {
 		reportErrorDB(_db);
 		return false;


jluebbe:
I'm not sure how to interpret the AppVeyor failure (changing the timeout shouldn't affect compilation).

jluebbe:
I've tried to rebuild my branch with AppVeyor and couldn't reproduce the error:
https://ci.appveyor.com/project/jluebbe/softhsmv2/builds/25248549

rijswijk:
> I've tried to rebuild my branch with AppVeyor and couldn't reproduce the error:
> https://ci.appveyor.com/project/jluebbe/softhsmv2/builds/25248549

It looks like the build timed out, we'll check and see if we can force a re-run.

jluebbe:
@rijswijk I've force-pushed this PR to trigger a rebuild, but now some Travis builds failed while the AppVeyor ones succeeded. I don't think that has anything to do with my change.

jluebbe:
@rijswijk, @halderen Is there anything left to do from my side?

halderen:
No there is nothing really needed.  We were (are) troubled with the builds failing but this really has nothing to do with this PR as we think.  Also I'd been looking at ways things could have some bad effects, but none found.

PR #481, opened 2019-05-28
Remove extra field iv_bits from CK_GCM_PARAMS

LudovicRousseau:
The PKCS#11 specification defines CK_GCM_PARAMS as:
```
typedef struct CK_GCM_PARAMS {
     CK_BYTE_PTR pIv;
     CK_ULONG ulIvLen;
     CK_BYTE_PTR pAAD;
     CK_ULONG ulAADLen;
     CK_ULONG ulTagBits;
   } CK_GCM_PARAMS;
```

### DIFF
--- a/src/lib/pkcs11/pkcs11.h
+++ b/src/lib/pkcs11/pkcs11.h
@@ -982,7 +982,6 @@ struct ck_aes_ctr_params {
 struct ck_gcm_params {
   unsigned char *iv_ptr;
   unsigned long iv_len;
-  unsigned long iv_bits;
   unsigned char *aad_ptr;
   unsigned long aad_len;
   unsigned long tag_bits;


LudovicRousseau:
The SoftHSM code compiles fine with my change. So the removed field iv_bits was never used.

See also https://www.cryptsoft.com/pkcs11doc/v230/structCK__GCM__PARAMS.html#_details

jariq:
You seem to be referring to PKCS#11 v2.30 specification but SoftHSM2 currently implements PKCS#11 v2.40 specification and the issue you are trying to solve is [one of many known issues of PKCS#11 v2.40 specification](https://github.com/Pkcs11Interop/Pkcs11Interop/issues/126#issuecomment-496687863). Moreover your change may break all libraries/apps that are based on official PKCS#11 v2.40 headers.

LudovicRousseau:
You are right.
The problem is in the unclear PKCS#11 specification.

xipki:
Just find out that the NSS header https://github.com/nss-dev/nss/blob/master/lib/util/pkcs11t.h does not have the field `iv_bits` in the `CK_GCM_PARAMS`.

rijswijk:
That is because NSS includes a very old version of the header files (looks like v2.11) that still comes from RSA Security, and that pre-dates v2.4 by more than a decade.

PR #482, opened 2019-05-30
Remove direct use of Exception class to work with modern Botan.

dewyatt:
Commit b909778 made this constructor protected.

This is a contribution from [Ribose Inc](https://www.ribose.com) (@riboseinc).


dewyatt:
Just a note that Jack let me know he reverted the change that made this constructor protected, but you may still want to merge this anyways to be *most* compatible.

rijswijk:
We'll review the change, it looks good and trivial at first glance.

halderen:
No direct opinion, pulling in and see how the build goes.

Issue #488, opened 2019-06-24
BUG: Definition of CK_GCM_PARAMS is NOT correct


xipki:
The definition of ck_gcm_params in `pkcs11.h` is NOT correct.

```
struct ck_gcm_params {
  unsigned char *iv_ptr;
  unsigned long iv_len;
  unsigned long iv_bits;
  unsigned char *aad_ptr;
  unsigned long aad_len;
  unsigned long tag_bits;
};
```
should be changed to

```
struct ck_gcm_params {
  unsigned char *iv_ptr;
  unsigned long iv_len;
  unsigned char *aad_ptr;
  unsigned long aad_len;
  unsigned long tag_bits;
};
```

The line  `unsigned long iv_bits;` should be removed.

The official Definition can be found in Section 2.12.3 of http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/os/pkcs11-curr-v2.40-os.html.

Note that the definition with iv_bits comes from the http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/pkcs11t.h. This, however, conflicts the spec.

jariq:
Duplicate of https://github.com/opendnssec/SoftHSMv2/pull/481.

rijswijk:
Closing, duplicate of #481 (thanks @jariq)

PR #489, opened 2019-06-24
Fixed bugs in issues 487 and 488.

xipki:


rijswijk:
We cannot merge this pull request; #488 is a duplicate of #481, for which we specified we would not fix this (it is an ambiguity in the v2.4 PKCS #11 spec). Will respond separately to the DSA issue you report.

PR #494, opened 2019-08-12
Check CK_ATTRIBUTE_PTR arguments for NULL_PTR

blukat29:
Add NULL_PTR checks for number of PKCS#11 functions that take CK_ATTRIBUTE_PTR as input arguments. Since NULL_PTR is allowed when `ulCount` is 0, the conditionals also check the `ulCount` (or similar) argument.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -1842,6 +1842,7 @@ CK_RV SoftHSM::C_SetAttributeValue(CK_SESSION_HANDLE hSession, CK_OBJECT_HANDLE 
 CK_RV SoftHSM::C_FindObjectsInit(CK_SESSION_HANDLE hSession, CK_ATTRIBUTE_PTR pTemplate, CK_ULONG ulCount)
 {
 	if (!isInitialised) return CKR_CRYPTOKI_NOT_INITIALIZED;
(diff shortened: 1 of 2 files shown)


rijswijk:
Thank you for this pull request, this seems like a useful addition to the code. Just to be sure: does this cover all cases where a ```NULL_PTR``` + ```ulCount == 0``` check could be performed?

blukat29:
I've visually inspected `SoftHSM::C_*` functions which takes CK_ULONG as input length argument. These are the functions I found relavant.
- C_CreateObject
- C_CopyObject
- C_GetAttributeValue
- C_SetAttributeValue
- C_FindObjectsInit
- C_GenerateKey
- C_GenerateKeyPair
- C_DeriveKey

Except the ones in the PR, most of them have simple NULL_PTR checks somewhere, either `C_*` function or inner implementation. I did not change them.

If you want me to add `ulCount != 0` to them, then I will work on it.

rijswijk:
Thanks for checking; if the other ```C_*``` functions have an inner check somewhere for ```NULL_PTR``` arguments for attribute templates (where relevant), then I think we can just merge the PR as is. I will ask my colleague to review and merge the PR.

halderen:
Useful check with no regression problems.

PR #500, opened 2019-08-28
Add support for C_WaitForSlotEvent

massey101:
C_WaitForSlotEvent is expected to be supplied as a pkcs11 function by
libraries and applications, such as nss. Since the slots are never
updated it's quite easy to implement.

Would you like me to create an issue and add an entry into NEWS?

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -7069,9 +7069,17 @@ CK_RV SoftHSM::C_CancelFunction(CK_SESSION_HANDLE hSession)
 }
 
 // Wait or poll for a slot event on the specified slot
-CK_RV SoftHSM::C_WaitForSlotEvent(CK_FLAGS /*flags*/, CK_SLOT_ID_PTR /*pSlot*/, CK_VOID_PTR /*pReserved*/)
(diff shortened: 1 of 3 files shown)


rijswijk:
Thanks, we'll update the news and check why the Travis build failed (probably some timeout)

halderen:
The travis build fails due to a crash in the botan library builds.  However the botan dependency pulled in from Ubuntu is version 1.10.12 from 2016.  There are much newer 1.10 versions available and the regular build check will use Botan version 2.  I recommend taking Travis with a grain of salt, we have other test platforms available which show no consistent errors.

Jakuje:
FYI, you can use Ubuntu 18.04 with travis already to avoid outdated dependencies:

https://docs.travis-ci.com/user/reference/bionic/

For this change request, I think the NSS for which was this change done is using particularly the blocking version of this call to avoid polling of all the slots in p11-kit as described in  https://github.com/OpenSC/OpenSC/issues/1771

I think this should also implement the blocking semantics per PKCS#11 specification:

>  If `C_WaitForSlotEvent` is called with the `CKF_DONT_BLOCK` flag clear in the flags argument, then the call behaves as above, except that it will block. That is, if no slot's event flag is set at the time of the call, `C_WaitForSlotEvent` will wait until some slot's event flag becomes set. If a thread of an application has a `C_WaitForSlotEvent` call blocking when another thread of that application calls `C_Finalize`, the `C_WaitForSlotEvent` call returns with the value `CKR_CRYPTOKI_NOT_INITIALIZED`.

PR #501, opened 2019-09-08
Remove support for obsolete/EOL versions of Botan

### LINKED ISSUE #493: Dropping support for Botan 1.10?
Botan upstream is considering removing public access to a number of headers starting in the next major release (expected ~2020) https://github.com/randombit/botan/pull/2061 adds warnings when these headers are included. In almost all cases, it is easy to avoid using these headers by instead using some other part of the public API.

Right now SoftHSM is affected by some of these header deprecations, largely because it creates objects like Botan::MD5 directly instead of going through the factory methods. Unfortunately, how that worked changed between 1.10 and 2.x so it is not possible to have a single code which works in both versions.

Would you consider dropping support for Botan < 2.0? All such versions are completely end of life since end of 2018. In addition Botan 2 has much better performance and side channel safety across the board, especially in RSA and ECDSA. And it appears most package distributions now ship some relatively recent Botan 2 version https://repology.org/project/botan/versions

I'd be willing to create a patch that removes support for Botan < 2.0.0 and also ensures that SoftHSM avoids any interface which is going to run into issues with Botan 3 onwards. This would also simplify SoftHSM code quite a bit, and validate that we are not going to remove anything that would in the future break SoftHSM.

Let me know if such a patch would be acceptable. Thanks.

randombit:
Also taking care to avoid headers which are now deprecated for public usage and which will be removed in a future major release, so this code  should work with any version of Botan 2 or (future) Botan 3.

Closes #493


randombit:
For AppVeyor build the prebuilt binaries in https://github.com/disig/SoftHSM2-AppVeyor need to be updated, or else an alternate source like vcpkg can be used.

PR #418, opened 2018-09-27
Set fields to NULL to avoid double free

bmurray7:
I encountered this issue while attempting to diagnose the root cause of #417. I followed the procedures described by @parasssh, but I made two changes
1. I used a PKCS11 URL for which there was no matching object
2. I used the newly added `-check` flag in `openssl pkey`

I was using the latest master branches of SoftHSMv2 and lbp11. For OpenSSL, I used v1.1.1.

I then ran the command, resulting in a segfault
```
#pkey -engine pkcs11 -inform engine -in 'token=slot_111111-id=%11%11'   -check
engine "pkcs11" set.
Format not recognized!
The key ID is not a valid PKCS#11 URI
The PKCS#11 URI format is defined by RFC7512
The legacy ENGINE_pkcs11 ID format is also still accepted for now
Format not recognized!
The key ID is not a valid PKCS#11 URI
The PKCS#11 URI format is defined by RFC7512
The legacy ENGINE_pkcs11 ID format is also still accepted for now
PKCS11_get_private_key returned NULL
cannot load key from engine
140737353732544:error:80065064:pkcs11 engine:ctx_load_key:invalid id:eng_back.c:659:
140737353732544:error:26096080:engine routines:ENGINE_load_private_key:failed loading private key:crypto/engine/eng_pkey.c:78:
unable to load key
Segmentation fault
```
I attached GDB and obtained a stacktrace:
```Program received signal SIGSEGV, Segmentation fault.
0x0000000000000101 in ?? ()
>>>>where
#0  0x0000000000000101 in ?? ()
#1  0x00007ffff69dc136 in SoftHSM::C_Finalize (this=0x6d80a0, pReserved=<optimized out>) at SoftHSM.cpp:547
#2  0x00007ffff69c3611 in C_Finalize (pReserved=0x0) at main.cpp:148
#3  0x00007ffff6c6dcb3 in pkcs11_CTX_unload (ctx=0x6d5cd0) at p11_load.c:149
#4  0x00007ffff6c71d3a in PKCS11_CTX_unload (ctx=0x6d5cd0) at p11_front.c:53
#5  0x00007ffff6c67114 in ctx_finish (ctx=0x6d28a0) at eng_back.c:359
#6  0x00007ffff6c6643a in engine_finish (engine=0x6d1ad0) at eng_front.c:159
#7  0x00007ffff7798ae9 in engine_unlocked_finish (e=0x6d1ad0, unlock_for_handlers=0) at crypto/engine/eng_init.c:61
[...]

Issue #417, opened 2018-09-21
SoftHSM + engine_pkcs11 + openssl fails with EC keys when compiled with openssl 1.1.1


parasssh:
See https://github.com/OpenSC/libp11/issues/249 for details. 
The issue is that when SoftHSM and libp11 is compiled with openssl-1.1.1, trying to access EC keys on the SoftHSM via openssl engine fails. But when compiled with openssl-1.0.2p, issue is not seen. 

Opened a issue here since I am not sure if this will be a bug with openssl or opensc or softhsm

bellgrim:
Was this issue resolved with the closed OpenSC issue and #418?

PR #503, opened 2019-09-13
Mark as not initialized in destructor

ansasaki:
This is a follow-up on #418 

I was investigating a crash similar to #417 which happened if a wrong PIN was provided when using the openssl + engine_pkcs11 + SoftHSM combination. The root of the problem is that the destructor for SoftHSM can be executed before the ``atexit()`` handlers installed by other libraries holding references to SoftHSM.

In one particular case, the engine_pkcs11 would call ``SoftHSM::C_CloseSession()`` when exiting, which would end in a crash in ``HandleManager::getSession()`` because the ``handleManager`` had been destroyed earlier.

The bug was reported as part of https://github.com/OpenSC/libp11/issues/273

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -403,6 +403,8 @@ SoftHSM::~SoftHSM()
 	mechanisms_table.clear();
 	supportedMechanisms.clear();
 
+	isInitialised = false;
+
 	resetMutexFactoryCallbacks();
 }
 


rijswijk:
Thanks for the PR, we'll review it and merge it if it passes all build tests.

ansasaki:
These failures in the CI seem to be unrelated with the change I proposed (I'm seeing failures when running the tests using the current develop branch).

Is there anything I can do?

ansasaki:
Hello @rijswijk, @halderen, could you please consider merging this before the new release?

Without this I get crashes in specific scenarios, specially when using OpenSSL + engine_pkcs11 + SoftHSM.

rijswijk:
Yes, certainly, thanks for reminding us, @halderen is actually collecting input for a release.

PR #506, opened 2019-10-09
Check key type in C_EncryptInit and C_DecryptInit

blukat29:
Add CK_KEY_TYPE checks at C_EncryptInit and C_DecryptInit just like in MacSignInit(). Fixes test cases on the way.


rijswijk:
Looks like a useful addition, thanks for your contribution. We will review the code and merge it after the review and CI checks.

PR #508, opened 2019-11-08
fix: set rpath for correct lib path resolution in conftest executables

### LINKED ISSUE #507: configure script does not consider OpenSSL library path in test executables
When providing a custom OpenSSL path by ``--with-openssl`` the configure script does not handle this path properly. That is, the path is considered when building the test executables ``conftest`` (function ``ac_fn_c_try_link``), however, not when running them (``ac_fn_c_try_run``).

On systems with native OpenSSL libraries, the configure script would in this case test the native shared library's capabilities, on systems without OpenSSL libraries the test would fail, and hence various OpenSSL specific options would be disabled, although the custom built OpenSSL library would provide the required functionality (e.g. P256, P384, P512, ED448, ED25519).

This behaviour possibly also applies to other dynamic libraries whose path can be specified by means of command line parameters.

0xa5a5:
OpenSSL and Botan are linked dynamically against test programs built by configure. When executing those test programs explicitly provided library paths should be considered as library search paths.

The following solutions were considered:
- Wrap AC_RUN_IFELSE and temporarily set LD_LIBRARY_PATH.
- Static linkage
- Use linker option to set the hard-coded run-time search path for conftest executables.

This PR implements the latter solution, which can be implemented with minimal changes in Botan and OpenSSL related m4 files.

Closes #507.


rijswijk:
Thanks for the quick PR, I'll ask for a review.

halderen:
The usage of -Wl,-rpath is a GNU-ism, which will break other linkers.  One would need to use libtool, or an equivalence in case of cmake.  The problem occurs when botan is placed outside of a default linker search path, as would be the case of any dependent library outside of the default path (e.g. openssl, cppunit).  If you install softhsm in the same location, the subsequent program that uses that would need such a fix too, which we cannot supply.  Therefor in case (not being uncommon) of undefault locations, use LD_LIBRARY_PATH instead, which can also be set during configure time.

PR #513, opened 2020-01-20
Add a configuration option to reset SoftHSM state on fork

ansasaki:
Add the configuration option "library.reset_on_fork".  When set to "true", will
make SoftHSM to reset its state on fork.

---
Was  "Add a configuration option to emulate hardware device"

Previous description:

Add the configuration option "slots.hardware".  When set to "true", will
make SoftHSM to set the CKF_HW_SLOT flag in the structure returned by
C_GetSlotInfo().

The added configuration option is useful for testing applications which
requires hardware slots and/or change behaviour for hardware slots.

Also reset the SoftHSM instance when a fork is detected and the
CKF_HW_SLOT flag is set. This emulates a common behaviour for
hardware devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ src/lib/session_mgr/test/sessionmgrtest
 src/lib/slot_mgr/test/slotmgrtest
 src/lib/test/p11test
 src/lib/test/softhsm2-alt.conf
+src/lib/test/softhsm2-reset-on-fork.conf
 src/lib/test/softhsm2-mech.conf
 src/lib/test/softhsm2.conf
 src/lib/test/tokens/64d6c3fe-1575-1736-1d26-5ccb28440ea7/
--- a/configure.ac
+++ b/configure.ac
@@ -215,6 +215,7 @@ AC_CONFIG_FILES([
 	src/lib/test/Makefile
 	src/lib/test/softhsm2.conf
 	src/lib/test/softhsm2-alt.conf
+	src/lib/test/softhsm2-reset-on-fork.conf
 	src/lib/test/softhsm2-mech.conf
 	src/lib/test/tokens/dummy
 	src/bin/Makefile
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -79,6 +79,12 @@
 #include <stdlib.h>
 #include <algorithm>
 
+#ifdef _WIN32
+#include <process.h>
+#else
+#include <unistd.h>
+#endif
+
 // Initialise the one-and-only instance
 
 #ifdef HAVE_CXX11
@@ -338,6 +344,12 @@ SoftHSM* SoftHSM::i()
 	if (!instance.get())
 	{
 		instance.reset(new SoftHSM());
+	} else if(instance->detectFork())
+	{
+		if (Configuration::i()->getBool("library.reset_on_fork", false))
+		{
+			instance.reset(new SoftHSM());
+		}
 	}
 
 	return instance.get();
@@ -360,6 +372,11 @@ SoftHSM::SoftHSM()
 	sessionManager = NULL;
 	handleManager = NULL;
 	resetMutexFactoryCallbacks();
+#ifdef _WIN32
+	forkID = _getpid();
+#else
+	forkID = getpid();
+#endif
 }
 
 // Destructor
@@ -12278,4 +12295,12 @@ bool SoftHSM::isMechanismPermitted(OSObject* key, CK_MECHANISM_PTR pMechanism)
 	}
 
 	return allowed.find(pMechanism->mechanism) != allowed.end();
+}
+
+bool SoftHSM::detectFork(void) {
+#ifdef _WIN32
+	return forkID != _getpid();
+#else
+	return forkID != getpid();
+#endif
 }
--- a/src/lib/SoftHSM.h
+++ b/src/lib/SoftHSM.h
@@ -193,6 +193,8 @@ private:
 	SessionManager* sessionManager;
 	HandleManager* handleManager;
 
+	int forkID;
+
 	// Encrypt/Decrypt variants
 	CK_RV SymEncryptInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechanism, CK_OBJECT_HANDLE hKey);
 	CK_RV AsymEncryptInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechanism, CK_OBJECT_HANDLE hKey);
@@ -478,5 +480,7 @@ private:
 
 	static bool isMechanismPermitted(OSObject* key, CK_MECHANISM_PTR pMechanism);
 	static void prepareSupportedMecahnisms(std::map<std::string, CK_MECHANISM_TYPE> &t);
+
+	bool detectFork(void);
 };
 
--- a/src/lib/common/Configuration.cpp
+++ b/src/lib/common/Configuration.cpp
@@ -49,6 +49,7 @@ const struct config Configuration::valid_config[] = {
 	{ "log.level",			CONFIG_TYPE_STRING },
 	{ "slots.removable",		CONFIG_TYPE_BOOL },
 	{ "slots.mechanisms",		CONFIG_TYPE_STRING },
+	{ "library.reset_on_fork",	CONFIG_TYPE_BOOL },
 	{ "",				CONFIG_TYPE_UNSUPPORTED }
 };
 
--- a/src/lib/common/softhsm2.conf.5.in
+++ b/src/lib/common/softhsm2.conf.5.in
@@ -82,6 +82,16 @@ slots.mechanisms = ALL
 .fi
 .RE
 .LP
+.SH LIBRARY.RESET_ON_FORK
+If set to true, the library will reset the state on fork.
+Default is false.
+.LP
+.RS
+.nf
+library.reset_on_fork = true
+.fi
+.RE
+.LP
 .SH ENVIRONMENT
 .TP
 SOFTHSM2_CONF
--- a/src/lib/common/softhsm2.conf.in
+++ b/src/lib/common/softhsm2.conf.in
@@ -11,3 +11,6 @@ slots.removable = false
 
 # Enable and disable PKCS#11 mechanisms using slots.mechanisms.
 slots.mechanisms = ALL
+
+# If the library should reset the state on fork
+library.reset_on_fork = false
--- a/src/lib/test/CMakeLists.txt
+++ b/src/lib/test/CMakeLists.txt
@@ -23,6 +23,7 @@ set(SOURCES p11test.cpp
             AsymWrapUnwrapTests.cpp
             TestsBase.cpp
             TestsNoPINInitBase.cpp
+            ForkTests.cpp
             ../common/log.cpp
             ../common/osmutex.cpp
             )
@@ -40,5 +41,6 @@ add_test(NAME ${PROJECT_NAME} COMMAND ${PROJECT_NAME}
 set(builddir ${PROJECT_BINARY_DIR})
 configure_file(softhsm2.conf.in softhsm2.conf)
 configure_file(softhsm2-alt.conf.in softhsm2-alt.conf)
+configure_file(softhsm2-reset-on-fork.conf.in softhsm2-reset-on-fork.conf)
 configure_file(softhsm2-mech.conf.in softhsm2-mech.conf)
 configure_file(tokens/dummy.in tokens/dummy)
--- a/src/lib/test/ForkTests.cpp
+++ b/src/lib/test/ForkTests.cpp
@@ -1,0 +1,138 @@
+/*
+ * Copyright (c) 2020 by Red Hat, Inc.
+ *
+ * Author: Anderson Toshiyuki Sasaki <ansasaki@redhat.com>
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+ * IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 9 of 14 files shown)


rijswijk:
Thank you for your PR; I was wondering if you could provide a description of the particular use case for which you implemented this functionality. Also, before merging, you should know that we are working on a project to support "real" hardware features in an upcoming SoftHSM release. One of your Red Hat colleagues (@letoams) is on the SoftHSM governance board and knows more about these plans, I suggest you also reach out to him.

ansasaki:
> Thank you for your PR; I was wondering if you could provide a description of the particular use case for which you implemented this functionality.

We had problems in the past with forking applications which behave well with SoftHSM, but would fail with real smart cards. When the application forks, the SoftHSM duplicates it's whole state, since everything is on memory. That is not the case with real smart cards.

With this feature, we can automate our test suite to try both behaviours and find these "bad" applications which can't handle de-initialization on fork. Without this, tests with real smart cards are required to detect such "bad" applications.

> Also, before merging, you should know that we are working on a project to support "real" hardware features in an upcoming SoftHSM release. One of your Red Hat colleagues (@letoams) is on the SoftHSM governance board and knows more about these plans, I suggest you also reach out to him.

Thank you for letting me know, I'll get in touch with him to know what his project is about. But I feel it is not what I need, but rather the opposite. What I want is to be able to test some hardware behaviour even when I don't have such hardware. I guess his work is about adding support for using hardware to store objects, or to encrypt the objects?

rijswijk:
> > Thank you for your PR; I was wondering if you could provide a description of the particular use case for which you implemented this functionality.
> 
> We had problems in the past with forking applications which behave well with SoftHSM, but would fail with real smart cards. When the application forks, the SoftHSM duplicates it's whole state, since everything is on memory. That is not the case with real smart cards.
> 
> With this feature, we can automate our test suite to try both behaviours and find these "bad" applications which can't handle de-initialization on fork. Without this, tests with real smart cards are required to detect such "bad" applications.

Thanks for explaining the use case, and what an odd behaviour; this almost has to be a deliberate implementation on the part of the smart card PKCS11 implementer, because on a normal fork you'd expect the full state including libraries and their state to be replicated. Maybe it is to do with the PC/SC session becoming invalid after ```fork()```?

> 
> > Also, before merging, you should know that we are working on a project to support "real" hardware features in an upcoming SoftHSM release. One of your Red Hat colleagues (@letoams) is on the SoftHSM governance board and knows more about these plans, I suggest you also reach out to him.
> 
> Thank you for letting me know, I'll get in touch with him to know what his project is about. But I feel it is not what I need, but rather the opposite. What I want is to be able to test some hardware behaviour even when I don't have such hardware. I guess his work is about adding support for using hardware to store objects, or to encrypt the objects?

No, you're right, it's not what you need, but it is a conflicting development, as it will -- at some point -- lead to a "legitimate" reason to use the ```CKF_HW_SLOT``` flag, but with different semantics.

I would like to ask if you would be willing to re-factor the PR to make this a compile-time feature that is turned off by default (e.g. call it ```--enable-softhw-test```). Then, if you are willing to do so, I'd like to bring this to the attention of the SoftHSM board to discuss it. My reasoning for this is that this is a very specific feature that supports a test use case, whereas work is ongoing on a potentially conflicting major feature release. I hope you understand. Please don't start implementing it quite yet if you say yes, because I'd like to run it past the board first to make sure they're "on board" (pun intended).

ansasaki:
> Thanks for explaining the use case, and what an odd behaviour; this almost has to be a deliberate implementation on the part of the smart card PKCS11 implementer, because on a normal fork you'd expect the full state including libraries and their state to be replicated. Maybe it is to do with the PC/SC session becoming invalid after `fork()`?

I think it is deliberate. Could be because of the PC/SC session, or could be that the module uses some process-to-session matching. In any case, this is a common behaviour among modules.

> I would like to ask if you would be willing to re-factor the PR to make this a compile-time feature that is turned off by default (e.g. call it `--enable-softhw-test`). Then, if you are willing to do so, I'd like to bring this to the attention of the SoftHSM board to discuss it. My reasoning for this is that this is a very specific feature that supports a test use case, whereas work is ongoing on a potentially conflicting major feature release. I hope you understand. Please don't start implementing it quite yet if you say yes, because I'd like to run it past the board first to make sure they're "on board" (pun intended).

I understand the reasoning. What if I added an compilation option to add an extra field to the state for extensions like this one? It could be something like `--enable-testing-extensions`. Then add a field, let's say 32 bits long, and use bit masks for different extensions.

Then I could add a configuration option like `extensions.deinit_on_fork` to set the bit for this behaviour. I'm proposing configuration options because I need a run-time switch.

Maybe it would be interesting for the future to have an extra field for SoftHSM extensions like this one.

rijswijk:
> > Thanks for explaining the use case, and what an odd behaviour; this almost has to be a deliberate implementation on the part of the smart card PKCS11 implementer, because on a normal fork you'd expect the full state including libraries and their state to be replicated. Maybe it is to do with the PC/SC session becoming invalid after `fork()`?
> 
> I think it is deliberate. Could be because of the PC/SC session, or could be that the module uses some process-to-session matching. In any case, this is a common behaviour among modules.
> 
> > I would like to ask if you would be willing to re-factor the PR to make this a compile-time feature that is turned off by default (e.g. call it `--enable-softhw-test`). Then, if you are willing to do so, I'd like to bring this to the attention of the SoftHSM board to discuss it. My reasoning for this is that this is a very specific feature that supports a test use case, whereas work is ongoing on a potentially conflicting major feature release. I hope you understand. Please don't start implementing it quite yet if you say yes, because I'd like to run it past the board first to make sure they're "on board" (pun intended).
> 
> I understand the reasoning. What if I added an compilation option to add an extra field to the state for extensions like this one? It could be something like `--enable-testing-extensions`. Then add a field, let's say 32 bits long, and use bit masks for different extensions.
> 
> Then I could add a configuration option like `extensions.deinit_on_fork` to set the bit for this behaviour. I'm proposing configuration options because I need a run-time switch.
> 
> Maybe it would be interesting for the future to have an extra field for SoftHSM extensions like this one.

I think that would be a good approach, I'm going to consult the SoftHSM board and will get back as soon as I hear from them.

rijswijk:
(for reference: have asked the SoftHSM board for feedback latest by Feb 1st)

jariq:
I've seen this behavior with quite a few commercial HSMs and also found it weird at first but then I found out that [PKCS#11 v2.20](https://github.com/Pkcs11Interop/PKCS11-SPECS/tree/master/v2.20) states in chapter 6.6.1 that child process should not keep the cryptoki nor login state of the parent:

> Consider a UNIX process **P** which becomes a Cryptoki application by calling `C_Initialize`, and then uses the `fork()` system call to create a child process **C**. Since **P** and **C** have separate address spaces (or will when one of them performs a write operation, if the operating system follows the copy-on-write paradigm), they are not part of the same application. Therefore, if **C** needs to use Cryptoki, it needs to perform its own `C_Initialize` call. Furthermore, if **C** needs to be logged into the token(s) that it will access via Cryptoki, it needs to log into them even if **P** already logged in, since **P** and **C** are completely separate applications.

So IMO compliant PKCS#11 implementation should treat forked process as a completely new process (all states need to be discarted).

ansasaki:
> So IMO compliant PKCS#11 implementation should treat forked process as a completely new process (all states need to be discarted).

Thank you for the information! I always thought that this behaviour was something like a de-facto standard adopted by the vendors. But it is stated in the specification.

Should we consider changing the default SoftHSM behaviour?

rijswijk:
Thanks @jariq that is useful input. I read it slightly differently, in that the text in the PKCS11 spec is prescriptive for applications rather than Cryptoki implementers, and that applications should expect the library to require a separate initialisation, but there is no mandated behaviour for token implementations. Admittedly, the text is ambiguous. Going forward I think a good solution would be to make this behaviour configurable, with the default being the current behaviour.

@ansasaki would you be willing to rework your code to do that? I would then also suggest removing the "hardware emulation" behaviour, since that is not really needed anymore (unless I am mistaken).

ansasaki:
> Thanks @jariq that is useful input. I read it slightly differently, in that the text in the PKCS11 spec is prescriptive for applications rather than Cryptoki implementers, and that applications should expect the library to require a separate initialisation, but there is no mandated behaviour for token implementations. Admittedly, the text is ambiguous. Going forward I think a good solution would be to make this behaviour configurable, with the default being the current behaviour.
> 
> @ansasaki would you be willing to rework your code to do that? I would then also suggest removing the "hardware emulation" behaviour, since that is not really needed anymore (unless I am mistaken).

Sure, I can change my PR to that direction, and drop the option to set the slot as a hardware slot. But I still need the run-time switch for that, which means I still need a configuration option and a way to store the option in the state.

Would the solution I described before (adding an extension field to the state) acceptable for this purpose?

rijswijk:
> > Thanks @jariq that is useful input. I read it slightly differently, in that the text in the PKCS11 spec is prescriptive for applications rather than Cryptoki implementers, and that applications should expect the library to require a separate initialisation, but there is no mandated behaviour for token implementations. Admittedly, the text is ambiguous. Going forward I think a good solution would be to make this behaviour configurable, with the default being the current behaviour.
> > @ansasaki would you be willing to rework your code to do that? I would then also suggest removing the "hardware emulation" behaviour, since that is not really needed anymore (unless I am mistaken).
> 
> Sure, I can change my PR to that direction, and drop the option to set the slot as a hardware slot. But I still need the run-time switch for that, which means I still need a configuration option and a way to store the option in the state.
> 
> Would the solution I described before (adding an extension field to the state) acceptable for this purpose?

I don't think it'd have to be an extension state, given @jariq's input this should be a regular feature. I suggest adding a configuration option ```library.reset_on_fork``` which, if set to ```true``` resets all state after a fork. The config option should be optional and the default should be ```false``` to keep the current behaviour as standard (so as not to inadvertently break anything that depends on it). Does that answer your question?

letoams:
On Wed, 22 Jan 2020, Roland van Rijswijk wrote:

> I don't think it'd have to be an extension state, given @jariq's input this should be a regular feature. I suggest adding
> a configuration option library.reset_on_fork which, if set to true resets all state after a fork. The config option
> should be optional and the default should be false to keep the current behaviour as standard (so as not to inadvertently
> break anything that depends on it). Does that answer your question?

If this option is only used for doing testing and shouldn't be set
during production, then the name library.reset_on_fork is not a
very good one. library.testing.reset_on_fork would be better.

For libreswan, all these options are passed to our daemon with an
--impair- prefix, making it very clear you doing something that
is really not standard and impairs your production quality.

Paul

ansasaki:
> I don't think it'd have to be an extension state, given @jariq's input this should be a regular feature. I suggest adding a configuration option `library.reset_on_fork` which, if set to `true` resets all state after a fork. The config option should be optional and the default should be `false` to keep the current behaviour as standard (so as not to inadvertently break anything that depends on it). Does that answer your question?

Yes it does, thank you.

> If this option is only used for doing testing and shouldn't be set during production, then the name library.reset_on_fork is not a very good one. library.testing.reset_on_fork would be better. For libreswan, all these options are passed to our daemon with an --impair- prefix, making it very clear you doing something that is really not standard and impairs your production quality. Paul

For me this is fine too as the purpose I want to use this option is for testing. But I wouldn't say that is not standard as it is a common behaviour among hardware devices. If an application fails when the option is set, it means the application would fail with many (maybe the majority of) hardware devices. 

I will change my PR with the given inputs.

rijswijk:
> > I don't think it'd have to be an extension state, given @jariq's input this should be a regular feature. I suggest adding a configuration option `library.reset_on_fork` which, if set to `true` resets all state after a fork. The config option should be optional and the default should be `false` to keep the current behaviour as standard (so as not to inadvertently break anything that depends on it). Does that answer your question?
> 
> Yes it does, thank you.
> 
> > If this option is only used for doing testing and shouldn't be set during production, then the name library.reset_on_fork is not a very good one. library.testing.reset_on_fork would be better. For libreswan, all these options are passed to our daemon with an --impair- prefix, making it very clear you doing something that is really not standard and impairs your production quality. Paul
> 
> For me this is fine too as the purpose I want to use this option is for testing. But I wouldn't say that is not standard as it is a common behaviour among hardware devices. If an application fails when the option is set, it means the application would fail with many (maybe the majority of) hardware devices.
> 
> I will change my PR with the given inputs.

I would also say it's not just for testing, so go ahead and make it a standard, albeit default "off" option. Thanks again!

ansasaki:
> I would also say it's not just for testing, so go ahead and make it a standard, albeit default "off" option. Thanks again!

Updated with the agreed behaviour.

@rijswijk Could you please check the new version?

rijswijk:
Thanks! I'm going to ask for a review and will then merge the PR.

ansasaki:
I pushed a small code style fix.

> There seems to be a small memory leak associated with the forking itself. Interestingly, there is about 2KB lost at the child process. How and where baffles me as the memory isn't leaked when faking the fork(), nor when using a memory sanitizer.

I tried to find this memory leak by running under valgrind, but I found just a small memory leak of 192 bytes. I'll open a new PR to fix it.

How are you detecting such memory leak?

rijswijk:
Thanks, and also thanks for the separate PR for this leak. 

One final heads up: I see you added a Red Hat copyright statement in the test code for the feature. While that block is license-compatible with the 2-clause BSD license that SoftHSMv2 has, I just want to make sure that you are aware that the SoftHSM project now lives under the Commons Conservancy, and that governance decisions about the project are made by the SoftHSM Governance Board (on which we currently have a Red Hat representative, @letoams).

ansasaki:
> Thanks, and also thanks for the separate PR for this leak.
> 
> One final heads up: I see you added a Red Hat copyright statement in the test code for the feature. While that block is license-compatible with the 2-clause BSD license that SoftHSMv2 has, I just want to make sure that you are aware that the SoftHSM project now lives under the Commons Conservancy, and that governance decisions about the project are made by the SoftHSM Governance Board (on which we currently have a Red Hat representative, @letoams).

The force of habit made me to put the Red Hat copyright statement. I believe it wouldn't be a problem to remove it if it is against your current licensing policy. Let me know if I should remove the statement.

rijswijk:
> > Thanks, and also thanks for the separate PR for this leak.
> > One final heads up: I see you added a Red Hat copyright statement in the test code for the feature. While that block is license-compatible with the 2-clause BSD license that SoftHSMv2 has, I just want to make sure that you are aware that the SoftHSM project now lives under the Commons Conservancy, and that governance decisions about the project are made by the SoftHSM Governance Board (on which we currently have a Red Hat representative, @letoams).
> 
> The force of habit made me to put the Red Hat copyright statement. I believe it wouldn't be a problem to remove it if it is against your current licensing policy. Let me know if I should remove the statement.

I don't believe it would be a problem, it's just to let you know that all code has been transferred to the SoftHSM project under the move to Commons Conservancy, and that if we merge this PR that will happen to this code as well.

halderen:
Thanks for the other fix leaking memory, but after some additional digging I've found the memory leak in prepareSupportedMecahnisms(),  Its indeed related to some memory that otherwise wouldn't pose a problem, but with forking() becomes a bit more prominent.  I found it by just repeated forking, where the child forks and the parent just exits.  You see the heap slowly increasing.
Since this leak isn't really due to this fix, I'll just merge in this fix and create a seperate PR for it.

Issue #323, opened 2017-06-04
EdDSA support


fxdupont:
For DNSSEC we need EdDSA support for Ed25519 and Ed448 curves. The first one is supported by a recently merged PR in github OpenSSL (1.1.1), there is nothing on the PKCS#11 side but it already has an ECC framework and extension mechanisms so I decided to experiment for DNSSEC usages (cf RFC 8080) and ISC Bind 9 (working)...

bellgrim:
Moving this to 2.5.0. EDDSA is still a draft in PKCS#11.

rijswijk:
I think it would be good to implement the draft spec from PKCS #11 in 2.5.0

bellgrim:
Fixed

jariq:
Can you please post link to mentioned "PKCS#11 EDDSA draft 4" ?

bellgrim:
https://www.oasis-open.org/committees/document.php?document_id=62198&wg_abbrev=pkcs11

PR #522, opened 2020-02-25
Fix advertised min and max mechanism sizes according to final PKCS#11 3.0 specification

Jakuje:
The section "2.3.10 Edwards Elliptic curve key pair generation" lists the key limits as following:

> For this mechanism, the only allowed values are 255 and 448 as RFC 8032 only defines curves of these two sizes.
> A Cryptoki implementation may support one or both of these curves and should set the ulMinKeySize and ulMaxKeySize fields accordingly.

[1] https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/csprd01/pkcs11-curr-v3.0-csprd01.html#_Toc10560876

### DIFF
--- a/src/lib/crypto/BotanEDDSA.cpp
+++ b/src/lib/crypto/BotanEDDSA.cpp
@@ -396,13 +396,13 @@ bool BotanEDDSA::deriveKey(SymmetricKey **ppSymmetricKey, PublicKey* publicKey, 
 unsigned long BotanEDDSA::getMinKeySize()
 {
 	// Only Ed25519 is supported
-	return 32*8;
+	return 255;
 }
 
 unsigned long BotanEDDSA::getMaxKeySize()
 {
 	// Only Ed25519 is supported
-	return 32*8;
+	return 255;
 }
 
 bool BotanEDDSA::reconstructKeyPair(AsymmetricKeyPair** ppKeyPair, ByteString& serialisedData)
--- a/src/lib/crypto/OSSLEDDSA.cpp
+++ b/src/lib/crypto/OSSLEDDSA.cpp
@@ -363,13 +363,13 @@ bool OSSLEDDSA::deriveKey(SymmetricKey **ppSymmetricKey, PublicKey* publicKey, P
 unsigned long OSSLEDDSA::getMinKeySize()
 {
 	// Ed25519 is supported
-	return 32*8;
+	return 255;
 }
 
 unsigned long OSSLEDDSA::getMaxKeySize()
 {
 	// Ed448 is supported
-	return 57*8;
+	return 448;
 }
 
 bool OSSLEDDSA::reconstructKeyPair(AsymmetricKeyPair** ppKeyPair, ByteString& serialisedData)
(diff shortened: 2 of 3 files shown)


rijswijk:
Thanks for contributing this. Quick question: did you check if the ```ulMinKeySize``` and ```ulMaxKeySize``` are not used somewhere else in the code for input validation if a key is imported (or something like that)? Just to make sure it doesn't break anything like that.

Jakuje:
> Quick question: did you check if the `ulMinKeySize` and `ulMaxKeySize` are not used somewhere else in the code for input validation if a key is imported (or something like that)? Just to make sure it doesn't break anything like that.

Yes, I did quick check and I did not notice these numbers anywhere in the source code. The number `57*8 = 456` is enough non-standard and I did not see it anywhere and nothing is checked against 256 in the `EDDSATests.cpp` either if I am right.

rijswijk:
> > Quick question: did you check if the `ulMinKeySize` and `ulMaxKeySize` are not used somewhere else in the code for input validation if a key is imported (or something like that)? Just to make sure it doesn't break anything like that.
> 
> Yes, I did quick check and I did not notice these numbers anywhere in the source code. The number `57*8 = 456` is enough non-standard and I did not see it anywhere and nothing is checked against 256 in the `EDDSATests.cpp` either if I am right.

OK, thanks, I'm going to ask for a review and to merge the PR.

halderen:
I can see it specified as such, but I'd like to understand it as the key size is 256 bits.  Also this change would need to be applied too to the OSSL layer.

Jakuje:
> I can see it specified as such, but I'd like to understand it as the key size is 256 bits. Also this change would need to be applied too to the OSSL layer.

The change is applied to both OSSL and Botan versions of the code.

The key is usually stored as a 256 bit key for convenience, but to my understanding, the first bit is always zero.

halderen:
I think we want to hold this change until the PKCS#11 has become approved.  I've verified that this change indeed does not cause any side effects.  The actual size (apart from some sanity checks) isn't used in determining the actual key size of buffer space needed.  The real key size for ED25519 and ED448 is determined by the specific algorithm used which is in another part of the template given.  The key size therefor for EC can be left out.
However we do worry that there will be a mismatch between programmer expectation and the value returned.  There will be implementations that will divide the returned value by 8 go get the expected key storage size needed.  RFCs on ED25519 mention the key size of 256 bits, the PKCS#11 is correct that there will be 255 bits relevant, but this may be less practical.  If the final PKCS#11 specification for 3.0 keeps this as is, we will probably adopt this.

Jakuje:
The 3.0 is approved for month or so:
https://www.oasis-open.org/news/announcements/four-pkcs-11-specifications-from-pkcs11-tc-approved-as-committee-specifications

The same statement was already in the previous versions, which were used to implement the EdDSA in SoftHSM in #323 so (see the last comment with link). Even though it might be less practical, applications can be using this max and min value to check if the pkcs11 module supports particular key size and if softhsm announces range 256-256 and application wants (correct) 255, it can wrongly assume softhsm does not support this key size, which was the issue I had when implementing support for this in OpenSC tools.

rijswijk:
@halderen can you merge this, since the 3.0 spec is final? There has been a request for this feature from someone else.

Jakuje:
Rebased on top of current develop branch.

Issue #523, opened 2020-02-25
EdDSA keygen not work according to final PKCS #11 3.0


Jakuje:
If I read the PKCS #11 3.0 specification correctly, the `EC_PARAMS` in the keygen template should list the DER encoding of curve name (PrintableString) for EdDSA keys. The template  in [1]  even adds an example how the ec params template should look like:

    CK_BYTE ecParams[] = {0x13, 0x0c, 0x65, 0x64, 0x77, 0x61, 0x72, 0x64, 0x73, 0x32, 0x35, 0x35, 0x31, 0x39};

But from my reading of the code in `src/lib/crypto/OSSLEDDSA.cpp`, the content is passed to few OpenSSL functions assuming it is really only OID and hoping we will get some NID from OpenSSL.

This probably needs to be updated to work with named curves according to the standard. Filling the issue now so it does not get lost, but I hope I will be able to work on this soon.

[1] https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/csprd01/pkcs11-curr-v3.0-csprd01.html#_Toc10560871

rijswijk:
Thanks, the contributions are much appreciated!

PR #526, opened 2020-02-26
 Adjust EDDSA code to return valid EC_PARAMS according to the final PKCS #11 3.0 specification

Jakuje:
The latest draft specifies that the EC_PARAMS should be DER encoding of PrintableString
and not OID.

https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/csprd01/pkcs11-curr-v3.0-csprd01.html#_Toc10560871

Fixes: #523


Jakuje:
The botan tests are still failing for some unrelated reasons. It passed for me locally on Fedora.

PR #528, opened 2020-02-26
Avoid creating duplicate mechanisms

Jakuje:
I noticed weird behavior when the library is loaded into longer-running
process, which issues several cycles of C_Initialize & C_Finalize,
where I saw exactly the same amount of duplicate items as the amount
of the calls to C_Initialize.

I was not able to replicated it in a testsuite, probably because of
different way how the memory is handled inside of C++ code and when
called from C code as a PKCS#11 library.

This fix is just making sure that the list is empty before filling it
with new mechanisms. Other possiblity would be to move the clean to
C_Finalize(), which is counterpart to the C_Initialize(), containing
the initialization of this list.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -596,6 +596,8 @@ CK_RV SoftHSM::C_Finalize(CK_VOID_PTR pReserved)
 
 	isInitialised = false;
 
+	supportedMechanisms.clear();
+
 	SoftHSM::reset();
 	return CKR_OK;
 }
@@ -777,6 +779,7 @@ void SoftHSM::prepareSupportedMecahnisms(std::map<std::string, CK_MECHANISM_TYPE
 	t["CKM_EDDSA"]			= CKM_EDDSA;
 #endif
 
+	supportedMechanisms.clear();
 	for (auto it = t.begin(); it != t.end(); ++it)
 	{
 		supportedMechanisms.push_back(it->second);


rijswijk:
Thanks for the PR.

It would actually make sense to have the explicit clean in both ```C_Initialize``` and in ```C_Finalize```. The behaviour you're seeing may be due to out-of-sync calling of both functions (so if the applications calls ```C_Initialize``` for a second time before calling ```C_Finalize``` for its first session. I'm typically a bit weary of accommodating badly behaved applications, but in this case it doesn't seem to hurt to add this fix.

One final question: did you check the call to clear the list does not leave any dangling allocations (all the items on the list are stack-allocated and have decent destructors)?

Jakuje:
> It would actually make sense to have the explicit clean in both `C_Initialize` and in `C_Finalize`.

Right. I can add it also there.

> The behaviour you're seeing may be due to out-of-sync calling of both functions (so if the applications calls `C_Initialize` for a second time before calling `C_Finalize` for its first session. I'm typically a bit weary of accommodating badly behaved applications, but in this case it doesn't seem to hurt to add this fix.

This should not be the case. I have a pkcs11-spy trace, where the C_Initialize and C_Finalize are properly matching (though they could have been called from different threads if it would make difference).

If the calls would not be matching and the C_Finalize would not be called before the next call to teh C_Initialize, we would get `CKR_CRYPTOKI_ALREADY_INITIALIZED` from the following calls, which would have been caught by the calling application (and which is not the case in the trace I had).

> One final question: did you check the call to clear the list does not leave any dangling allocations (all the items on the list are stack-allocated and have decent destructors)?

This list has only members (or copies?)  from `mechanisms_table`, but I am really not a C++ developer so I might not be getting this right so review of handling these structures would be helpful.

rijswijk:
> > It would actually make sense to have the explicit clean in both `C_Initialize` and in `C_Finalize`.
> 
> Right. I can add it also there.

Please do, just to make sure we catch the potential bad behaviour discussed.

> > The behaviour you're seeing may be due to out-of-sync calling of both functions (so if the applications calls `C_Initialize` for a second time before calling `C_Finalize` for its first session. I'm typically a bit weary of accommodating badly behaved applications, but in this case it doesn't seem to hurt to add this fix.
> 
> This should not be the case. I have a pkcs11-spy trace, where the C_Initialize and C_Finalize are properly matching (though they could have been called from different threads if it would make difference).
> 
> If the calls would not be matching and the C_Finalize would not be called before the next call to teh C_Initialize, we would get `CKR_CRYPTOKI_ALREADY_INITIALIZED` from the following calls, which would have been caught by the calling application (and which is not the case in the trace I had).

OK, thanks for the explanation, sounds like you're right and the behaviour I suggested is not occurring.

> > One final question: did you check the call to clear the list does not leave any dangling allocations (all the items on the list are stack-allocated and have decent destructors)?
> 
> This list has only members (or copies?) from `mechanisms_table`, but I am really not a C++ developer so I might not be getting this right so review of handling these structures would be helpful.

I checked the code, it is safe to clear the list, there are no allocations that would become dangling (everything happens on the stack).

Jakuje:
Added also to `C_FInalize`.

rijswijk:
Thanks, I've asked for a review

PR #530, opened 2020-03-03
Appveyor build platform update

halderen:


jariq:
Hello @halderen, do I understand it correctly that these change removed the dependency on prebuilt items that were fetched from https://github.com/disig/SoftHSM2-AppVeyor repository? If yes, I'd like to proceed and delete that repo. Are you OK with that?

halderen:
Go ahead.  I had the suggestion to use vcpkg instead as this uses appveyor provided builds.  So yes we don't kneed them anymore.  Thanks for the provided support.

\Berry

PR #539, opened 2020-03-26
Resolve issue #538

### LINKED ISSUE #538: Constexpr assignment operators in PublicKey and PrivateKey
I have few questions/issues about these copy assignments introduced recently. 

https://github.com/opendnssec/SoftHSMv2/blob/e4c954e96ae5b625ca800b78f9fee5e2cc453423/src/lib/crypto/PrivateKey.h#L71
https://github.com/opendnssec/SoftHSMv2/blob/e4c954e96ae5b625ca800b78f9fee5e2cc453423/src/lib/crypto/PublicKey.h#L64

1) It does not make much sense to define an assignment operator `const` function - you should not assign into a constant variable.
2) The implementation also seems weird. You would not expect an assignment operator to do nothing.
3) According to the standard, assignment operator can be `constexpr` only if all the types involved are `LiteralType`, which the keys are not - they have virtual destructor.

Is there some reason behind it? If not, I suggest to change this to plain:

```c++
PrivateKey& operator= (const PrivateKey&) = default;
```

Or, if the purpose was to silence warnings but disable assignment, then:

```c++
PrivateKey& operator= (const PrivateKey&) = delete;
```

Both should work in all compilers, these are common snippets.

Aearsis:
Fixes #538.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -1080,6 +1080,7 @@ CK_RV SoftHSM::C_GetMechanismInfo(CK_SLOT_ID slotID, CK_MECHANISM_TYPE type, CK_
 #endif
 		case CKM_DES3_CBC:
 			pInfo->flags = CKF_WRAP;
+			// falls through
 		case CKM_DES3_ECB:
 		case CKM_DES3_CBC_PAD:
 			// Key size is not in use
@@ -1100,6 +1101,7 @@ CK_RV SoftHSM::C_GetMechanismInfo(CK_SLOT_ID slotID, CK_MECHANISM_TYPE type, CK_
 			break;
 		case CKM_AES_CBC:
 			pInfo->flags = CKF_WRAP;
+			// falls through
 		case CKM_AES_ECB:
 		case CKM_AES_CBC_PAD:
 		case CKM_AES_CTR:
--- a/src/lib/crypto/AsymmetricKeyPair.h
+++ b/src/lib/crypto/AsymmetricKeyPair.h
@@ -42,14 +42,6 @@
 class AsymmetricKeyPair : public Serialisable
 {
 public:
-	// Base constructors
-	AsymmetricKeyPair() { }
-
-	AsymmetricKeyPair(const AsymmetricKeyPair& /*in*/) { }
-
-	// Destructor
-	virtual ~AsymmetricKeyPair() { }
-
 	// Return the public key
 	virtual PublicKey* getPublicKey() = 0;
 	virtual const PublicKey* getConstPublicKey() const = 0;
--- a/src/lib/crypto/AsymmetricParameters.h
+++ b/src/lib/crypto/AsymmetricParameters.h
@@ -40,19 +40,8 @@
 class AsymmetricParameters : public Serialisable
 {
 public:
-	// Base constructors
-	AsymmetricParameters() { }
-
-	AsymmetricParameters(const AsymmetricParameters& /*in*/) { }
-
-	// Destructor
-	virtual ~AsymmetricParameters() { }
-
 	// Check if it is of the given type
 	virtual bool areOfType(const char* type) = 0;
-
-	// Serialisation
-	virtual ByteString serialise() const = 0;
 };
 
 #endif // !_SOFTHSM_V2_ASYMMETRICPARAMETERS_H
--- a/src/lib/crypto/PrivateKey.h
+++ b/src/lib/crypto/PrivateKey.h
@@ -41,14 +41,6 @@
 class PrivateKey : public Serialisable
 {
 public:
-	// Base constructors
-	PrivateKey() { }
-
-	PrivateKey(const PrivateKey& in);
-
-	// Destructor
-	virtual ~PrivateKey() { }
-
 	// Check if the private key is of the given type
 	virtual bool isOfType(const char* inType) = 0;
 
@@ -58,17 +50,11 @@ public:
 	// Get the output length
 	virtual unsigned long getOutputLength() const = 0;
 
-	// Serialisation
-	virtual ByteString serialise() const = 0;
-
 	// Encode into PKCS#8 DER
 	virtual ByteString PKCS8Encode() = 0;
 
 	// Decode from PKCS#8 BER
 	virtual bool PKCS8Decode(const ByteString& ber) = 0;
-
-	// Assignment
-	constexpr const PrivateKey& operator=(const PrivateKey&) const { return *this; }
 };
 
 #endif // !_SOFTHSM_V2_PRIVATEKEY_H
--- a/src/lib/crypto/PublicKey.h
+++ b/src/lib/crypto/PublicKey.h
@@ -40,11 +40,6 @@
 class PublicKey : public Serialisable
 {
 public:
-	// Base constructors
-	PublicKey() { }
-
-	PublicKey(const PublicKey& /*in*/) { }
-
 	// Destructor
 	virtual ~PublicKey() { }
 
@@ -56,12 +51,6 @@ public:
 
 	// Get the output length
 	virtual unsigned long getOutputLength() const = 0;
-
-	// Serialisation
-	virtual ByteString serialise() const = 0;
-
-	// Assignment
-	constexpr const PublicKey& operator=(const PublicKey&) const { return *this; }
 };
 
 #endif // !_SOFTHSM_V2_PUBLICKEY_H
--- a/src/lib/crypto/SymmetricKey.cpp
+++ b/src/lib/crypto/SymmetricKey.cpp
@@ -42,12 +42,6 @@ SymmetricKey::SymmetricKey(size_t inBitLen /* = 0 */)
 	bitLen = inBitLen;
 }
 
-SymmetricKey::SymmetricKey(const SymmetricKey& in)
-{
-	keyData = in.keyData;
-	bitLen = in.bitLen;
-}
-
 // Set the key
 bool SymmetricKey::setKeyBits(const ByteString& keybits)
 {
--- a/src/lib/crypto/SymmetricKey.h
+++ b/src/lib/crypto/SymmetricKey.h
@@ -40,13 +40,8 @@
 class SymmetricKey : public Serialisable
 {
 public:
-	// Base constructors
-	SymmetricKey(size_t inBitLen = 0);
-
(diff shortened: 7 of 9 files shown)


rijswijk:
Thanks for the PR, I will ask @halderen to include it in the final 2.6.0, this looks like a cleaner way to fix the gcc10 issues.

iay:
I'm not sure how important really old compilation environments are, but two data points:

* This PR allows compilation under Mac OS X 10.9, which current `master` does not. Huzzah!

* Unfortunately, this PR appears to produce some _other_ errors under Mac OS 10.7 (Lion) with Xcode 4.6.3:

```
                 from AESKey.cpp:34:
./../common/Serialisable.h:48: error: invalid pure specifier (only `= 0' is allowed) before ‘;’ token
./../common/Serialisable.h:49: error: invalid pure specifier (only `= 0' is allowed) before ‘;’ token
./../common/Serialisable.h:50: error: expected ‘,’ or ‘...’ before ‘&&’ token
./../common/Serialisable.h:50: error: invalid pure specifier (only `= 0' is allowed) before ‘;’ token
./../common/Serialisable.h:50: error: invalid constructor; you probably meant ‘Serialisable (const Serialisable&)’
./../common/Serialisable.h:52: error: invalid pure specifier (only `= 0' is allowed) before ‘;’ token
./../common/Serialisable.h:53: error: expected ‘,’ or ‘...’ before ‘&&’ token
./../common/Serialisable.h:53: error: invalid pure specifier (only `= 0' is allowed) before ‘;’ token
./../common/Serialisable.h:48: error: initializer specified for non-virtual method ‘Serialisable::Serialisable()’
./../common/Serialisable.h:49: error: initializer specified for non-virtual method ‘Serialisable::Serialisable(const Serialisable&)’
./../common/Serialisable.h:52: error: initializer specified for non-virtual method ‘Serialisable& Serialisable::operator=(const Serialisable&)’
./../common/Serialisable.h:53: error: initializer specified for non-virtual method ‘Serialisable& Serialisable::operator=(Serialisable)’
./../common/Serialisable.h:50: error: cannot declare parameter ‘<anonymous>’ to be of abstract type ‘Serialisable’
./../common/Serialisable.h:42: note:   because the following virtual functions are pure within ‘Serialisable’:
./../common/Serialisable.h:45: note: 	virtual ByteString Serialisable::serialise() const
./../common/Serialisable.h:53: error: cannot declare parameter ‘<anonymous>’ to be of abstract type ‘Serialisable’
./../common/Serialisable.h:42: note:   since type ‘Serialisable’ has pure virtual functions
```

rijswijk:
Wow, 10.7 that is quite ancient ;-). I'm guessing this is also due to not supporting newer C++ features, would it be fair to say we don't support such an old system? Or do you think time should be invested in that?

Aearsis:
Well, first of all, we should define a version of C++ against which we compile. The `&&`s which baffle the compiler are C++11 move semantics, and C++11 is, well, used since ~2011 :)

We can simply drop the move constructors & assignments. Actually, to make use of the optimization, the code using Serialisables would have to be changed. For now, I would remove them, as they give no immediate benefit. But in the near future, I would like to see (at least) `-std=c++11` and some refactoring to speed things up :)

Aearsis:
After some fiddling with `-pedantic`, it seems that some sort of support for C++11 is already required - namely variadic macros and `auto` as a placeholder type specifier (SoftHSM.cpp). Also, GNU extension (`hh` printf modifier) is used in the `dump` utility.

The new version should be almost compilable in C++98 mode, with excetpions listed above. @iay , can you please test it? I do not happen to have a Mac around :)

iay:
> Wow, 10.7 that is quite ancient ;-). I'm guessing this is also due to not supporting newer C++ features, would it be fair to say we don't support such an old system? Or do you think time should be invested in that?

Yes, 10.7 (Lion) is not recent (last release of it was October 2012). It comes up because I'm working on the MacPorts Portfile, and that project supports everything back to 10.6 (!). Whether you want to put effort into supporting older systems is of course your choice, I'm just feeding back observations you may find relevant (and may affect the ability to compile on other old systems too, not just Macs).

I'm pretty new to the whole MacPorts system, but it looks like it's possible in principle to use a different compiler (e.g., some version of gcc) rather than the default. I'm not sure it's worth the effort though.

iay:
> The new version should be almost compilable in C++98 mode, with excetpions listed above. @iay , can you please test it? I do not happen to have a Mac around :)

I don't have a 10.7 Mac around either, but I did set up a VM for testing. Your latest PR gets a lot further in the compilation, stopping instead with this:

```
SoftHSM.cpp: In member function ‘void SoftHSM::prepareSupportedMecahnisms(std::map<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int, std::less<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int> > >&)’:
SoftHSM.cpp:784: error: ISO C++ forbids declaration of ‘it’ with no type
SoftHSM.cpp:784: error: cannot convert ‘std::_Rb_tree_iterator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int> >’ to ‘int’ in initialization
SoftHSM.cpp:784: error: no match for ‘operator!=’ in ‘it != ((std::map<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int, std::less<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int> > >*)t)->std::map<_Key, _Tp, _Compare, _Alloc>::end [with _Key = std::basic_string<char, std::char_traits<char>, std::allocator<char> >, _Tp = long unsigned int, _Compare = std::less<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, _Alloc = std::allocator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long unsigned int> >]()’
SoftHSM.cpp:786: error: base operand of ‘->’ is not a pointer
SoftHSM.cpp: In member function ‘CK_RV SoftHSM::C_GetMechanismList(CK_SLOT_ID, CK_MECHANISM_TYPE*, CK_ULONG*)’:
SoftHSM.cpp:856: error: ISO C++ forbids declaration of ‘it’ with no type
SoftHSM.cpp:856: error: ‘class std::list<long unsigned int, std::allocator<long unsigned int> >’ has no member named ‘cbegin’
SoftHSM.cpp:857: error: ‘class std::list<long unsigned int, std::allocator<long unsigned int> >’ has no member named ‘cend’
SoftHSM.cpp:859: error: invalid type argument of ‘unary *’
SoftHSM.cpp: In member function ‘bool SoftHSM::isMechanismPermitted(OSObject*, _CK_MECHANISM*)’:
SoftHSM.cpp:12427: error: ISO C++ forbids declaration of ‘it’ with no type
SoftHSM.cpp:12427: error: cannot convert ‘std::_List_iterator<long unsigned int>’ to ‘int’ in initialization
SoftHSM.cpp:12428: error: no match for ‘operator==’ in ‘it == mechs. std::list<_Tp, _Alloc>::end [with _Tp = long unsigned int, _Alloc = std::allocator<long unsigned int>]()’
```

I think these were some of the exceptions you were expecting?

A bit of guessing (I haven't programmed in C++ for real for 20 years) and I made this:

```
diff --git a/src/lib/SoftHSM.cpp b/src/lib/SoftHSM.cpp
index 4520af5..329be0d 100644
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -781,7 +781,7 @@ void SoftHSM::prepareSupportedMecahnisms(std::map<std::string, CK_MECHANISM_TYPE
 #endif
 
        supportedMechanisms.clear();
-       for (auto it = t.begin(); it != t.end(); ++it)
+       for (auto std::map<std::string, CK_MECHANISM_TYPE>::iterator it = t.begin(); it != t.end(); ++it)
        {
                supportedMechanisms.push_back(it->second);
        }
@@ -853,8 +853,8 @@ CK_RV SoftHSM::C_GetMechanismList(CK_SLOT_ID slotID, CK_MECHANISM_TYPE_PTR pMech
        *pulCount = nrSupportedMechanisms;
 
        int i = 0;
-       auto it = supportedMechanisms.cbegin();
-       for (; it != supportedMechanisms.cend(); it++, i++)
+       auto std::list<CK_MECHANISM_TYPE>::iterator it = supportedMechanisms.begin();
+       for (; it != supportedMechanisms.end(); it++, i++)
        {
                pMechanismList[i] = *it;
        }
@@ -12424,7 +12424,7 @@ bool SoftHSM::isMechanismPermitted(OSObject* key, CK_MECHANISM_PTR pMechanism)
 {
        std::list<CK_MECHANISM_TYPE> mechs = supportedMechanisms;
        /* First check if the algorithm is enabled in the global configuration */
-       auto it = std::find(mechs.begin(), mechs.end(), pMechanism->mechanism);
+       auto std::list<CK_MECHANISM_TYPE>::iterator it = std::find(mechs.begin(), mechs.end(), pMechanism->mechanism);
        if (it == mechs.end())
                return false;
```

That does seem to install and work at a basic level (I was able initialise a token, make a keypair, get openssl to make a self-signed cert for that and then import it).

`make check` doesn't work, though. _Probably_ more C++11 things (e.g., type_traits), but again my knowledge is too ancient to be able to tell easily.

All just FYI of course; I don't know whether this brings you close enough to supporting pre-C++11 compilers to want to try and support them, and I'm sure the world will survive if you don't decide to do that and instead declare C++11 your baseline. I agree that it would be a good idea to make that call explicit, whichever way you go.

rijswijk:
While there hasn't been an explicit decision, I think it would make sense to standardise on C++11 as a minimum requirement, it feels like it would be a stretch to maintain pre-C++11 compatibility.

iay:
I found an (as far as I can tell, completely undocumented) Portfile option to tell the MacPorts system that the port requires C++11. If I select that on a Mac OS X 10.7 system, it downloads and uses a completely different compiler and it looks like everything works (including testing, if I make another dependency tweak in the Portfile to make sure pkg-config is available).

I'll do a quick double-check that this seems to hold after this PR has been merged, then come back and make the Portfile changes once 2.6.1 is out.

Issue #403, opened 2018-06-27
Replace RFC4880 PBKDF by something more modern


rijswijk:
For the next version of SoftHSM, it is worth investigating what would be a worthwhile replacement for the somewhat dated RFC 4880 inspired PBKDF that we currently use to something more modern. For example, scrypt, as specified in RFC 7914, but it may be worth exploring other options as well.

Note: replacing the PBKDF will require some form of token versioning to support transitioning existing tokens.

PR #540, opened 2020-03-27
Implement C_CopyObject for DB backend

Aearsis:
Formerly there were no way to list all attributes, as they are scattered in different tables. This PR implements it using a view, through which all attributes can be seen. This view is then used for attribute enumeration, allowing C_CopyObject to work.

The view is added in `DBObject::migrateTables()`, which is called when opening existing token. It should be compatible with tokens created in previous version.

If we did not insist on minimalizing changes, it might be even better to throw all attributes in a single table - SQLite is fine with varying types of table cells.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -1618,6 +1618,7 @@ CK_RV SoftHSM::C_CopyObject(CK_SESSION_HANDLE hSession, CK_OBJECT_HANDLE hObject
 	{
 		if (!object->attributeExists(attrType))
 		{
+			WARNING_MSG("Attribute 0x%lx does not exist even though it was reported as next.");
 			rv = CKR_FUNCTION_FAILED;
 			break;
 		}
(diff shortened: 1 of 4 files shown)


rijswijk:
Ignore the CI warning, something appears broken there (it is on our radar).

Thanks for the PR, this seems like a somewhat more major change, any objections to scheduling this for merging after we release 2.6.0 and tagging it for a follow-up release?

Aearsis:
No objections at all, in fact I'm using this code for a few months and held it for the next release window myself (there were multiple requests to make a release and I did not want to postpone the release further).

After 2.6.0 we can discuss and develop a more generic approach to token versioning, from which more issues will profit (#403). I do not consider these commits absolutely final, as the approach of attempting migration in every single opening is not perfect.

halderen:
I don't think we should include this in 2.6.1 but in 2.7.0.  This change will perform a database change.  Not only should minor patch releases contain no schema changes (or even significant functional changes), but moreover this makes it impossible to downgrade to the previous patch version, or easily migrate a token storage from one installation to another (or back).  This is used and will be unexpected between minor patch versions.
I'd like the contribution, so I DO want to pull it in, just not in 2.6.1 but perhaps we can plan for a 2.7.0 to be soonish.

Aearsis:
Actually, the change should be both forward and backwards compatible (though only forward compatibility was tested, and just once :) ). The view is created automatically, and is ignored by older versions (AFAIK).

Also, from my POV no new feature is added - just that now C_CopyObject is broken, and can be fixed in a patch release. The db schema change is unfortunate, but I can't do better than this.

As I said though, no pressure from my side - we already need to use patched version of SoftHSM.

jschlyter:

PR #542, opened 2020-04-04
Get OpenSSL Ed448/X448 working

dev-aaront-org:
This puts the finishing touches on the OpenSSL Ed448/X448 implementation. I'm able to successfully run BIND's Ed448 DNSSEC tests with these changes.

### DIFF
--- a/src/lib/crypto/OSSLEDDSA.cpp
+++ b/src/lib/crypto/OSSLEDDSA.cpp
@@ -368,7 +368,7 @@ unsigned long OSSLEDDSA::getMinKeySize()
 
 unsigned long OSSLEDDSA::getMaxKeySize()
 {
-	// Ed448 will be supported
+	// Ed448 is supported
 	return 57*8;
 }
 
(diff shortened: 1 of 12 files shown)


src/lib/crypto/OSSLEDDSA.cpp
```diff
@@ -368,7 +368,7 @@ unsigned long OSSLEDDSA::getMinKeySize()
 
 unsigned long OSSLEDDSA::getMaxKeySize()
 {
-	// Ed448 will be supported
+	// Ed448 is supported
 	return 57*8;
```
oerdnj, line 372: Should you use `ED448_KEYLEN` here?

rijswijk:
Thanks for the review @oerdnj 

Going to assign this to @halderen for a review of the test and a merge if OK, may take some time due to limited working hours due to "the crisis".

src/lib/crypto/OSSLEDDSA.cpp
a-a-ron-t: Both the min and max values will be changed with PR #522, so I'll leave them as-is for this PR.

Issue #548, opened 2020-05-04
Use after free on ENGINE_finish(rdrand_engine).


dwmw2:
Commit caec0c7ea83ca has caused SoftHSM to stop exiting cleanly when running the `auth-pkcs11` test from OpenConnect self-tests. Instead of exiting, it just stops:

```
#0  __pthread_rwlock_wrlock_full (abstime=0x0, rwlock=0x555555cbb320) at pthread_rwlock_common.c:917
#1  __GI___pthread_rwlock_wrlock (rwlock=0x555555cbb320) at pthread_rwlock_wrlock.c:27
#2  0x00007ffff6e49e7d in CRYPTO_THREAD_write_lock (lock=<optimized out>) at crypto/threads_pthread.c:78
#3  0x00007ffff6daf745 in ENGINE_finish (e=0x555555b8af90) at crypto/engine/eng_init.c:101
#4  0x00007fffe99a54d7 in OSSLCryptoFactory::~OSSLCryptoFactory (this=0x55555583a1b0, __in_chrg=<optimized out>) at OSSLCryptoFactory.cpp:240
#5  0x00007fffe99a5559 in OSSLCryptoFactory::~OSSLCryptoFactory (this=0x55555583a1b0, __in_chrg=<optimized out>) at OSSLCryptoFactory.cpp:227
#6  0x00007ffff7a12680 in __run_exit_handlers (status=1, listp=0x7ffff7b98738 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true)
    at exit.c:108
#7  0x00007ffff7a127c0 in __GI_exit (status=<optimized out>) at exit.c:139
#8  0x0000555555559878 in main ()
```

This appears to be true for the GnuTLS build of OpenConnect but *not* the OpenSSL build.

dwmw2:
https://bugzilla.redhat.com/show_bug.cgi?id=1831086

dwmw2:
Valgrind says...
```
==228434== Invalid read of size 4
==228434==    at 0x54D9316: __pthread_rwlock_wrlock_full (pthread_rwlock_common.c:581)
==228434==    by 0x54D9316: pthread_rwlock_wrlock (pthread_rwlock_wrlock.c:27)
==228434==    by 0x58ADE7C: CRYPTO_THREAD_write_lock (threads_pthread.c:78)
==228434==    by 0x5813744: ENGINE_finish (eng_init.c:101)
[...]

PR #550, opened 2020-05-12
Issue #548: Don't clean up engines after OpenSSL has already shut down

dwmw2:
As of 1.1.0, OpenSSL registers its own atexit() handler to call
OPENSSL_cleanup(). If our own code subsequently tries to, for example,
unreference an ENGINE, then it'll crash or deadlock with a use after
free.

Fix it by registering a callback with OPENSSL_atexit() to be called when
OPENSSL_cleanup() is called. It sets a flag which prevents any further
touching of OpenSSL objects — which would otherwise happen fairly much
immediately thereafter when our own OSSLCryptoFactory destructor gets
called by the C++ runtime's own atexit() handler.

Fixes: #548


rijswijk:
Thanks for the PR, looks good to me, assigning for review and merge.

PR #551, opened 2020-05-13
Fix OPENSSL_cleanup() detection without using our own atexit() handler

dwmw2:
We can't register our own atexit() or OPENSSL_atexit() handler because
there's no way to unregister it when the SoftHSM DSO is unloaded. This
causes the crash reported at https://bugzilla.redhat.com/1831086#c8

Instead of using that method to set a flag showing that OPENSSL_cleanup()
has occurred, instead test directly by calling OPENSSL_init_crypto() for
something that *would* do nothing, but will fail if OPENSSL_cleanup()
has indeed been run already.

Fixes: c2cc0652b4 "Issue #548: Don't clean up engines after OpenSSL
                   has already shut down"

### DIFF
--- a/src/lib/crypto/OSSLCryptoFactory.cpp
+++ b/src/lib/crypto/OSSLCryptoFactory.cpp
@@ -77,7 +77,6 @@ bool OSSLCryptoFactory::FipsSelfTestStatus = false;
 
 static unsigned nlocks;
 static Mutex** locks;
-static bool ossl_shutdown;
 
 // Mutex callback
 void lock_callback(int mode, int n, const char* file, int line)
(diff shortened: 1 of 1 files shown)


dwmw2:
Note outstanding request for testing, at https://bugzilla.redhat.com/show_bug.cgi?id=1831086#c11

dwmw2:
Those CI failures don't look like they're mine...

dwmw2:
> Note outstanding request for testing, at https://bugzilla.redhat.com/show_bug.cgi?id=1831086#c11

Note confirmation of fix from user, at Note outstanding request for testing, at https://bugzilla.redhat.com/show_bug.cgi?id=1831086#c12

dwmw2:
No change other than commit date; just trying to prod the CI as some of those failures look transient.

dwmw2:
OK, the travis-ci build is now OK, and appveyor has been broken in master for a while. Ignoring that.

rijswijk:
Will ask @halderen for another review and to merge it if OK.

src/lib/crypto/OSSLCryptoFactory.cpp
```diff
@@ -250,7 +226,21 @@ OSSLCryptoFactory::OSSLCryptoFactory()
...
+	// OpenSSL 1.1.0+ will register an atexit() handler to run
+	// OPENSSL_cleanup(). If that has already happened we must
+	// not attempt to free any ENGINEs because they'll already
+	// have been destroyed and the use-after-free would cause
+	// a deadlock or crash.
+	//
+	// Detect that situation because reinitialisation will fail
+	// after OPENSSL_cleanup() has run.
+	(void)ERR_set_mark();
+	ossl_shutdown = !OPENSSL_init_crypto(OPENSSL_INIT_ENGINE_RDRAND, NULL);
```
nomis, line 241: It's possible to compile OpenSSL without RDRAND support.

PR #564, opened 2020-07-28
Test CI Appveyor

halderen:


halderen:
Current problem is that the getopt library won't properly integrate

jschlyter:
Appveyor removed.

PR #566, opened 2020-08-10
Add "objectstore.umask" configuration option for file/directory creation

### LINKED ISSUE #560: Allow file modes to be configured
#40 removed read permission from group/other for SoftHSM tokens

This makes it difficult to backup these files because no groups have access  to the files, even if there are additional default ACLs on the directory allowing other groups to access them.

It needs to be possible to configure a umask of `0027` instead of always using `0077`.

nomis:
This defaults to 0077 to preserve the existing behaviour. It can be set to
0027 to allow users in the same group to backup the token files.

It does not attempt to modify the process umask (because this is a library)
so it can only apply in addition to whatever umask the process has.

Fixes #560 after #40 restricted all new files to the current user.


rijswijk:
Thanks for the PR! I performed a first review, looks good at first glance, requesting a review from @halderen and asking him to merge if he also thinks the code is OK.

nomis:
Watch out for the `File` constructor when reviewing because C++ allows the umask argument to be passed a `bool`.

halderen:
Nice contribution.  Although most times the shell umask is used, I could see the this be overridden.
A generic comment I have is that all options in the configuration file are more-or-less mandatory at the moment, causing warnings if they're not explicitly set.  I need (in a separately pull request) to change this.

PR #576, opened 2020-09-11
Fix race condition between C_FindObjects and object creation (issue 573)

### LINKED ISSUE #573: Concurrency issue
Hello!

We've been using SoftHSM2 for testing  - thank you very much for this library, it's been very useful to us - but recently we've started getting random failures that seem concurrency-related.

Our tests do various operations, but most of them rely on `C_GenerateKeyPair` or `C_CreateObject` to begin with. These are the two that ultimately fail (with `CKR_GENERAL_ERROR` or `CKR_TEMPLATE_INCONSISTENT`) - however which call to generate/create fails is pretty much random. The failed tests pass when run independently or even when rerunning the same call immediately after a failure. They only fail when run in a multi-threaded way. Within the service we create a new PKCS11 session for each operation and that session is only used in one thread. The library is initialized with `CKF_OS_LOCKING_OK`.

We generally use RSA key pairs, and from syslogs I could see that the failure usually comes from `P11KeyObj::init`, `P11PublicKeyObj::init` or `P11PrivateKeyObj::init`.

Digging deeper and adding some logging, I found that the error surfaces from [`ObjectFile::setAttribute`](https://github.com/opendnssec/SoftHSMv2/blob/develop/src/lib/object_store/ObjectFile.cpp#L208) - `valid` ends up being false after the call to `store()`. The strange thing is that while logging at all possible return points of [`ObjectFile::store`](https://github.com/opendnssec/SoftHSMv2/blob/develop/src/lib/object_store/ObjectFile.cpp#L661) I could never see `valid` turning false...

I'm not really sure what else to try here, or whether this is because of how we handle multi-threading or some bug somewhere else in the stack. 

Just as a note, we access SoftHSM2 through the [`pkcs11`](https://crates.io/crates/pkcs11) Rust crate.

rijswijk:
This likely fixes #573

### DIFF
--- a/src/lib/object_store/ObjectFile.cpp
+++ b/src/lib/object_store/ObjectFile.cpp
@@ -707,8 +707,6 @@ void ObjectFile::store(bool isCommit /* = false */)
 		}
 	}
 
-	objectFile.flush();
-
 	valid = true;
 }
 
--- a/src/lib/object_store/ObjectFile.cpp
+++ b/src/lib/object_store/ObjectFile.cpp
@@ -322,7 +322,7 @@ void ObjectFile::refresh(bool isFirstTime /* = false */)
 	{
 		DEBUG_MSG("Object %s is empty", path.c_str());
 
-		valid = false;
+        objectFile.unlock();
 
 		return;
 	}
@@ -706,6 +706,8 @@ void ObjectFile::store(bool isCommit /* = false */)
 			return;
 		}
 	}
+
+    objectFile.flush();
 
 	valid = true;
 }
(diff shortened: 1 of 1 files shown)


nomis:
I've put a generic "flush before unlock" into File in #578, because none of the buffered reads or writes are flushed before unlocking, so the call to flush() will be unnecessary here.

rijswijk:
@nomis thanks for #578 and this note, I will wait for feedback from @ionut-arm and merge both PRs after positive feedback.

hug-dev:
Thanks for the change! Ionut will come back on Wednesday 16th. The changes look as planned so I am sure it would be fine to merge if you wanted to do that earlier 😃

ionut-arm:
Hi! Apologies for the delay - I've tested again and it seems the stress test still fails, but while before it was failing mainly with `CKR_GENERAL_ERROR`, now it fails with `CKR_TEMPLATE_INCONSISTENT`. I'll try to identify where the problem is coming from

ionut-arm:
Hey! Sorry for such a late reply, I managed to do some checks both locally and in the CI and it seems the latest change fixes our problems! 🎉 

If you're ok with it, we're all in for it to be merged.

PR #583, opened 2020-10-30
Support different OAEP modes using OpenSSL EVP interface

### LINKED ISSUE #474: Support OAEP mechanisms different than SHA1
The current version supports only SHA1 hash and SHA1-based MGF1 in the OAEP decryption mechanism, which is not very future proof.

This is more like known issue for visibility of the missing functionality. If there will be nobody interested in working on that, I might eventually take this.

borancar:
Using EVP interface, we can support different OAEP padding modes, such as OAEP-SHA224, OAEP-SHA256, OAEP-SHA384, OAEP-SHA512. This is my attempt at plumbing the support in - let me know if I'm doing it right. Fixes https://github.com/opendnssec/SoftHSMv2/issues/474.

## Summary by CodeRabbit

- New Features
  - Added support for RSA-OAEP with SHA-1, SHA-224, SHA-256, SHA-384, and SHA-512 across encrypt/decrypt and key wrap/unwrap.
  - OAEP mechanisms are now recognized for key wrapping.

- Improvements
  - More robust RSA encryption/decryption with proper input size validation and clearer error reporting.
  - Consistency checks ensure OAEP MGF matches the selected hash algorithm.

- Documentation
  - Updated minimum OpenSSL dependency to 1.0.2.


borancar:
I've tested decryption across all of them. Could someone help me test encryption?

borancar:
Implement all OAEP tests and they all pass now. The only workaround is not to use key size of 1024 for OAEP-SHA512.

rijswijk:
Thanks for this contribution; once you think it is ready for review, can you create a pull request so we can review the code?

rijswijk:
Had a quick look at the code, looks good at first glance. To bump the OpenSSL version, you should also update `m4/acx_crypto_backend.m4`.

borancar:
Done, thanks!

m4/acx_crypto_backend.m4
```diff
@@ -65,9 +65,9 @@ AC_DEFUN([ACX_CRYPTO_BACKEND],[
 		AC_MSG_RESULT(OpenSSL)
 
 		if test "x${enable_fips}" = "xyes"; then
-			ACX_OPENSSL(1,0,1)
+			ACX_OPENSSL(1,0,2)
 		else
-			ACX_OPENSSL(1,0,0)
+			ACX_OPENSSL(1,0,2)
```
realmfoo, line 72: If-branch and else-branch are the same.

borancar: And advice on version for FIPS compliance?

realmfoo: According to [FIPS-140](https://www.openssl.org/docs/fips.html) openssl version should be 1.0.2, but you are changing openssl version in both branches to 1.0.2. Indeed for FIPS it should stay at version 1.0.2.

It's odd that two branches uses the same version. May be we should add a note that version of OpenSSL for FIPS compatibility should stay exactly at version 1.0.2?

But I guess there is an issue with openssl version checking for FIPS. In acx_openssl.m4 an openssl version should be [at least](https://github.com/opendnssec/SoftHSMv2/blame/9f63ccc0529894b546a2b18c03197f7e30aab6ae/m4/acx_openssl.m4#L43) as specified version in ACX_OPENSSL. [OpenSSL FIPS notes](https://www.openssl.org/docs/fips.html) that "Neither validation will work with any release other than 1.0.2".

borancar: So guess it makes sense to keep it as-is, and the else branch will increase later as needed?

realmfoo: Yep. And it makes sense to add a comment why it has else branch and when it should be increased.

borancar: Done

jschlyter:

src/lib/SoftHSM.cpp
```diff
@@ -6247,11 +6300,51 @@ CK_RV SoftHSM::WrapKeyAsym
...
+				case CKM_SHA_1:
+					mech = AsymMech::RSA_PKCS_OAEP_SHA1;
+					expectedMgf = CKG_MGF1_SHA1;
+					// PKCS#11 2.40 draft 2 section 2.1.8: input length <= k-2-2hashLen
+					if (keydata.size() > modulus_length - 2 - 2 * 160 / 8)
+						return CKR_KEY_SIZE_RANGE;
+					break;
+				case CKM_SHA224:
+					mech = AsymMech::RSA_PKCS_OAEP_SHA224;
+					expectedMgf = CKG_MGF1_SHA1;
```
antoinelochet, line 6313: This (and the other `expectedMgf` below for the other hash functions) is not the same as the other `expectedMgf` is the modification above.
It seems weird.
I think it should be `CKG_MGF1_SHA224` which would make more sense.

antoinelochet:
SoftHSM does not compile with Botan backend.

src/lib/SoftHSM.cpp
```diff
@@ -12392,14 +12514,22 @@ CK_RV SoftHSM::MechParamCheckRSAPKCSOAEP(CK_MECHANISM_PTR pMechanism)
...
 	}
-	if (params->mgf != CKG_MGF1_SHA1)
+	if (params->mgf != CKG_MGF1_SHA1 &&
+	    params->mgf != CKG_MGF1_SHA224 &&
+	    params->mgf != CKG_MGF1_SHA256 &&
+	    params->mgf != CKG_MGF1_SHA384 &&
+	    params->mgf != CKG_MGF1_SHA512)
 	{
-		ERROR_MSG("mgf must be CKG_MGF1_SHA1");
+		ERROR_MSG("mgf must be onf of: CKG_MGF1_SHA1, CKM_MGF1_SHA224, CKM_MGF1_SHA256, CKM_MGF1_SHA384, CKM_MGF1_SHA512");
```
LawrenceHunter:
```suggestion
		ERROR_MSG("mgf must be one of: CKG_MGF1_SHA1, CKM_MGF1_SHA224, CKM_MGF1_SHA256, CKM_MGF1_SHA384, CKM_MGF1_SHA512");
```

src/lib/SoftHSM.cpp
borancar: What about this @antoinelochet?

bukka:
This was done in https://github.com/softhsm/SoftHSMv2/commit/de25233b8b9b6564d5a9ca56fc2d6e4d31ddb7eb

Issue #589, opened 2020-12-23
EDDSA signature failing - OSSLEDDSA.cpp(90): EDDSA sign init failed (0x060A109E) - looking for a message digest?


keldonin:
When trying to perform signature using a valid ED25519 private key generated on the token, calls to `C_Sign()` fail with a `CKR_GENERAL_ERROR`.

```
18: C_SignInit
>> 2020-12-23 10:30:25.007
[in] hSession = 0x1
pMechanism->type=CKM_EDDSA
[in] hKey = 0x2
<< 2020-12-23 10:30:25.008
Returned:  0 CKR_OK

19: C_Sign
>> 2020-12-23 10:30:25.008
[in] hSession = 0x1
[in] pData[ulDataLen] 0000000801a7d6e0 / 68
    00000000  30 42 02 01 00 30 0F 31 0D 30 0B 06 03 55 04 03  0B...0.1.0...U..
    00000010  0C 04 74 65 73 74 30 2A 30 05 06 03 2B 65 70 03  ..test0*0...+ep.
    00000020  21 00 00 50 0E 42 50 CB 2E 59 C6 75 A8 15 DB EF  !..P.BP..Y.u....
    00000030  9A 33 1F A4 B5 49 D9 0D 8D 81 F9 61 2A E7 09 AB  .3...I.....a*...
    00000040  23 7E A0 00                                      #~..
<< 2020-12-23 10:30:25.008
Returned:  5 CKR_GENERAL_ERROR
```
When setting library verbosity level to DEBUG, it yields the following:
```
OSSLEDDSA.cpp(90): EDDSA sign init failed (0x060A109E)
 ```

The OpenSSL code refers to OpenSSL EVP library, function `EVP_F_DO_SIGVER_INIT`, i.e. within `do_sigver_init()` from `crypto/evp/m_sigver.c` and is related to not finding a default digest.

`C_SigInit()` is invoked with no `CK_EDDSA_PARAMS` structure, assuming that SoftHSM implements the `PureEdDSA` variant of the algorithm.

Possibly a bug?

rijswijk:
The call to `C_SignInit` does not require parameters (see also the testcases in `src/lib/test/SignVerifyTests.cpp`). Running `make check` should exercise this function, and on my test system (Ubuntu 20.04 LTS) the tests pass without problems. What environment are you running your test on?

keldonin:
[...]

PR #590, opened 2020-12-23
fix for issue #589: specify `EVP_md_null()` as hashing algorithm for EdDSA sig and vfy

keldonin:
This patch hopefully fixes issue #589. I have tested for signature, using ED25519 and ED448 private keys. I couldn't validate for verify, but this should be trivial - maybe part of the test suite.

Regards,

### DIFF
--- a/src/lib/crypto/OSSLEDDSA.cpp
+++ b/src/lib/crypto/OSSLEDDSA.cpp
@@ -85,7 +85,7 @@ bool OSSLEDDSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
 	signature.resize(len);
 	memset(&signature[0], 0, len);
 	EVP_MD_CTX* ctx = EVP_MD_CTX_new();
-	if (!EVP_DigestSignInit(ctx, NULL, NULL, NULL, pkey))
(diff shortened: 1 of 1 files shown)


keldonin:
AppVeyor builds failed, but not because of this patch. There seems to be a more fundamental issue with the pipeline.

rijswijk:
The discussion in the OpenSC project (referenced above) seems to suggests using `EVP_md_null()` isn't necessary and also might not be the right approach. Please have a look at my question under #589.

dengert:
With OpenSC in  https://github.com/OpenSC/OpenSC/pull/2190 we were looking for an md, that would collect the input, not do hash, but present the total data collected as the hash. This would allow for ECDSA (and even RSA) to not have to treat the raw  mechanisms differently from the hashed mechanisms. But it looks like openssl/crypto/evp/m_null.c - EVP_md_null, just discards any data sent to it.

keldonin:
As explained in issue #589, this is a side effect of customizing EDDSA methods in OpenSSL. Withdrawing the PR.
Thanks for the review and suggestions!

PR #593, opened 2021-01-03
Fixing tests of other PKCS#11 implementations.

larssilven:
The unit tests is a very good tool to test other implementations.
These commits will change some tests so that testing of other implementations will have fewer failures.

### DIFF
--- a/src/lib/test/ForkTests.cpp
+++ b/src/lib/test/ForkTests.cpp
@@ -90,7 +90,7 @@ void ForkTests::testFork()
 	rv = CRYPTOKI_F_PTR( C_Finalize(NULL_PTR) );
 	CPPUNIT_ASSERT(rv == CKR_OK);
 }
-#ifndef P11M
+#ifndef P11_SHARED_LIBRARY
 void ForkTests::testResetOnFork()
 {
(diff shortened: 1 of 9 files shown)


rijswijk:
I see no problems with these additions, but I think it would be good if the `#define P11M` would have a bit more of a descriptive name (what does the "M" stand for?), and if you'd add a flag to `configure.ac` to set this `#define` (where the default is that it is disabled). Could you make those changes to the PR?

rijswijk:
Thanks, @halderen can you merge?

halderen (approved):
Reviewed. I assume other pkcs#11 libraries have other demands, so it might not be generic enough.  One particular point of interest was the change to the call to C_WaitForSlotEvent.  SoftHSMv2 does not assume the slotPtr the be a non-null pointer.  The PKCS#11 specification isn't really specific on this subject (see section 5.5) so one could say a null pointer might be allowed.  However in my reading a non-null pointer for an application program would be preferred, whereas pkcs#11 implementators should allow for a null pointer.  But since some won't, it does not hurt changing the test to this end as the specs might hint towards not using a null pointer.

PR #599, opened 2021-01-27
ANSI X9.63 key derivation functions for CKM_ECDH1_DERIVE

Aearsis:
Implemented in `SoftHSM.cpp`, so it works with both OpenSSL and Botan backends. The added tests are kind of elementary, I haven't found any official test vectors. But I have checked the results against doing the SHA's by hand, and against OpenSSL's internal implementation.

It would be awesome if someone could check the outputs with another PKCS#11 implementation.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -10343,6 +10343,7 @@ CK_RV SoftHSM::deriveECDH
 	CK_BBOOL isPrivate)
 {
 	*phKey = CK_INVALID_HANDLE;
+	HashAlgorithm* kdfAlgorithm = NULL;
 
 	if ((pMechanism->pParameter == NULL_PTR) ||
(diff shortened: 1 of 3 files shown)


rijswijk:
Thanks for the contribution, much appreciated! Code looks good to me, I'm going to discuss it internally, we are having some challenges with the review process due to corona-related pressure on time available to collaborators, so bear with us for some time (apologies in advance).

antoinelochet:
Hello @Aearsis,
Sorry to revive this conversation this long after the fact.
I have taken your patch for my needs but it seems incoherent compared to CKD_NULL regarding to the truncating.
While for CKD_NULL and the existing implementation, the method `ByteString::split` is used, you use `ByteString::resize`.
PKCS#11 specifies: "The truncation removes bytes from the leading end of the secret value."
As you, I could not find any test vectors either.
Maybe some in @rijswijk or someone in the tam could chip in ?

jschlyter:

PR #606, opened 2021-02-10
InfoTests::testGetMechanismNegativeListConfig() not working when testing

larssilven:
another shared library than SoftHSM.


rijswijk:
Makes sense, thanks for the fix.

PR #611, opened 2021-02-21
softhsm2-util: add --show-config parameter

alonbl:
Enables access to build configuration.

Currently implemented the `default-pkcs11-lib` variable to print the default
PKCS#11 library path.

Useful for applications which are using the softhsm2-util and require to
find the provider in filesystem which is used by the utility as there is no
standard process to locate the library.


rijswijk:
This looks like a useful feature, thanks for the PR. @halderen I suggest merging this.

alonbl:
Thanks!

I also updated the man page.

alonbl:
ping?

PR #612, opened 2021-02-21
Support import certificate using softhsm2-util

alonbl:
The softhsm2-util already support importing keys, why not also import
certificates?

Useful for test scripts that require both keys and certificates.

Add --import-type <type> parameter, depreciate the --aes parameter.

### DIFF
--- a/src/bin/util/softhsm2-util-botan.cpp
+++ b/src/bin/util/softhsm2-util-botan.cpp
@@ -47,6 +47,9 @@
 #include <botan/bigint.h>
 #include <botan/der_enc.h>
 #include <botan/oids.h>
+#include <botan/der_enc.h>
+#include <botan/x509cert.h>
+#include <botan/x509_dn.h>
 
 // Init Botan
 void crypto_init()
@@ -207,6 +210,64 @@ int crypto_import_key_pair
 
 	delete pkey;
 	return result;
+}
+
+int crypto_import_certificate
+(
+	CK_SESSION_HANDLE hSession,
+	char* filePath,
+	char* label,
+	char* objID,
+	size_t objIDLen
+)
+{
+	auto _vector_ptr = [](std::vector<uint8_t> &v) {
+		return v.size() == 0 ? NULL : &v.front();
+	};
+
+	Botan::X509_Certificate cert(filePath);
+	std::vector<uint8_t> blob;
+	std::vector<uint8_t> subject;
+	std::vector<uint8_t> issuer;
+	std::vector<uint8_t> serial;
+
+	Botan::DER_Encoder blob_encoder(blob);
+	Botan::DER_Encoder subject_encoder(subject);
+	Botan::DER_Encoder issuer_encoder(issuer);
+	Botan::DER_Encoder serial_encoder(serial);
+
+	cert.encode_into(blob_encoder);
+	cert.subject_dn().encode_into(subject_encoder);
+	cert.issuer_dn().encode_into(issuer_encoder);
+	serial_encoder.encode(Botan::BigInt::decode(cert.serial_number()));
+
+	CK_OBJECT_CLASS certClass = CKO_CERTIFICATE;
+	CK_CERTIFICATE_TYPE certType = CKC_X_509;
+	CK_BBOOL ckFalse = CK_FALSE, ckTrue = CK_TRUE;
+	CK_ATTRIBUTE certTemplate[] = {
+		{ CKA_CLASS,            &certClass,          	sizeof(certClass) },
+		{ CKA_CERTIFICATE_TYPE, &certType,           	sizeof(certType) },
+		{ CKA_LABEL,            label,               	strlen(label) },
+		{ CKA_ID,               objID,               	objIDLen },
+		{ CKA_TOKEN,            &ckTrue,            	sizeof(ckTrue) },
+		{ CKA_PRIVATE,		&ckFalse,		sizeof(ckFalse) },
+		{ CKA_VALUE,            _vector_ptr(blob),    	(CK_ULONG)blob.size() },
+		{ CKA_SUBJECT,          _vector_ptr(subject), 	(CK_ULONG)subject.size() },
+		{ CKA_ISSUER,           _vector_ptr(issuer),  	(CK_ULONG)issuer.size() },
+		{ CKA_SERIAL_NUMBER,    _vector_ptr(serial),  	(CK_ULONG)serial.size() }
+	};
+
+	CK_OBJECT_HANDLE hCert;
+	CK_RV rv = p11->C_CreateObject(hSession, certTemplate, 10, &hCert);
+	if (rv != CKR_OK)
+	{
+		fprintf(stderr, "ERROR: Could not save the certificate in the token.\n");
+		return 1;
+	}
+
+	printf("The certificate has been imported.\n");
+
+	return 0;
 }
 
 // Read the key from file
--- a/src/bin/util/softhsm2-util-ossl.cpp
+++ b/src/bin/util/softhsm2-util-ossl.cpp
@@ -218,6 +218,151 @@ int crypto_import_key_pair
 	return result;
 }
 
+int crypto_import_certificate
+(
+	CK_SESSION_HANDLE hSession,
+	char* filePath,
+	char* label,
+	char* objID,
+	size_t objIDLen
+)
+{
+	BIO* in = NULL;
+	X509* x509 = NULL;
+	CK_BYTE_PTR blob = NULL;
+	CK_BYTE_PTR subject = NULL;
+	CK_BYTE_PTR issuer = NULL;
+	CK_BYTE_PTR serial = NULL;
+	CK_BYTE_PTR p;
+	int blobSize;
+	int subjectSize;
+	int issuerSize;
+	int serialSize;
+	int ret = 1;
+
+	if (!(in = BIO_new_file(filePath, "rb")))
+	{
+		fprintf(stderr, "ERROR: Could open the PKCS#8 file: %s\n", filePath);
+		goto cleanup;
+	}
+
+	if ((x509 = PEM_read_bio_X509(in, NULL, NULL, NULL)) == NULL)
+	{
+		fprintf(stderr, "ERROR: Could not read the certificate file: %s\n", filePath);
+		goto cleanup;
+	}
+
+	blobSize = i2d_X509(x509, NULL);
+	subjectSize = i2d_X509_NAME(X509_get_subject_name(x509), NULL);
+	issuerSize = i2d_X509_NAME(X509_get_issuer_name(x509), NULL);
+	serialSize = i2d_ASN1_INTEGER(X509_get_serialNumber(x509), NULL);
+
+	if
+	(
+		blobSize < 0 ||
+		subjectSize < 0 ||
+		issuerSize < 0 ||
+		serialSize < 0
+	)
+	{
+		fprintf(stderr, "ERROR: Could not convert certificate to DER.\n");
+		goto cleanup;
+	}
+
+	if (blobSize > 0)
+	{
+		if ((blob = (CK_BYTE_PTR)malloc(blobSize)) == NULL)
+		{
+			fprintf(stderr, "ERROR: Could not allocate memory.\n");
+			goto cleanup;
+		}
+		p = blob;
+		blobSize = i2d_X509(x509, &p);
+	}
+	if (subjectSize > 0) {
+		if ((subject = (CK_BYTE_PTR)malloc(subjectSize)) == NULL)
+		{
+			fprintf(stderr, "ERROR: Could not allocate memory.\n");
+			goto cleanup;
+		}
+		p = subject;
+		subjectSize = i2d_X509_NAME(X509_get_subject_name(x509), &p);
+	}
+	if (issuerSize > 0)
+	{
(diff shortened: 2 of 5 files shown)


rijswijk:
Thank you for your contribution, I have some things to consider before merging:

Certificate import may be slightly more complex, since certificates are often associated with keys. Different applications link 
keys and certificates in different ways (e.g. one example might be to have matching `CKA_ID` attributes, but other options also exist). I'm not entirely sure whether adding this type of support to `softhsm2-util` is the right way forward, it might be too narrowly scoped to a particular use case.

@halderen your opinion would be welcome.

If we do decide to proceed with this, I think the following changes/features would be necessary:

- Don't do an immediate deprecation of ``--import``, instead, clearly mark it as deprecated in the documentation (usage, man page) and tag it for future removal (while you still support it, it is immediately deprecated without documentation)
- Support Botan as well (to keep the two builds in sync)

alonbl:
Hi,

Thank you for the review. I altered the implementation and added `--import-type` instead of deprecating the `--import`, I hope you like it better.

For the matching between key and certificate, as far as I know the best practice is to match the `CKA_ID` or read the entire token and match based on public key. The option to even having a basic import is significant for the use case of using the provider as test framework. If an application has special needs it can always perform the import by its own.

I have been around for years in PKCS#11 domain and never saw a setup in which the CKA_ID is not used for matching.

If you ACK I will update the man page and try to hack the botan as well.

Regards,
Alon

alonbl:
The botan part was actually simpler than I thought, so I applied the botan part and the man part.

rijswijk:
Thanks, looks good, one small request: the PR is still essentially deprecating the current mode of use of `--import` (which essentially assumes it's always a key pair unless `--aes` is specified). Furthermore, it looks like you removed the documentation for `--aes` from the manual page while the option is still silently supported. To really make it a "nice" deprecation, I suggest the following:

* If `--aes` is specified, print a warning to `stderr` that says something like "Warning: use of --aes is deprecated, this flag will be removed in a future version"
* If `--import` is specified without `--import-type`, print a warning to `stderr` that says something like "Warning: use of `--import` without specifying the object type with `--import-type` is deprecated, assuming `"keypair"`; future versions will give an error in this case."
* Clear flag the deprecated behaviour in the manual page

Again, thanks for your contribution, I think with the changes above we'd be good to go for a review by @halderen and a merge.

alonbl:
Hi,

I left the —aes in the man page, just removed it from the first line. And
added depreciation notice in parameter usage, so I do not think anything is
missing... just less endorsement for the parameter. :)

The past addition of the —aes was very confusing, I only realized it’s
usage when reading the code.

For the stderr messages, I do not like it, it may break existing scripts.
As long as we support backward compatibility, I thing the risk of breakage
due to output is wrong. I believe documentation should be sufficient.

What do you think?

Thanks,
Alon

rijswijk:
I think I agree, I'm going to ask @halderen to review and merge. Thanks!

alonbl:
ping?

alonbl:
Hi @halderen,
Can you please also checkout this one?
Thanks!

davydsantos:
+1 on the feature. This can be accomplished by using opensc/pkcs11-tool but it would be easier to rely on one single tool for both tasks. Another possible feature could be to allow importing a p12/pfx directly without the need of breaking it up in 2 different files and it would handle everything internally.

alonbl:
Rebased.

Hi @halderen,
Can you please also checkout this one? I will be glad to accept any feedback.
Thanks!

alonbl:
Hi @halderen,
Do you have time to review this?
Thanks.

alonbl:
ping?
I see almost no activity in project, any chance to merge it? feature was acked and change was reviewed.
thanks!

alonbl:
Hello @halderen, can you please look at this patch? It would be great if we can use the softhsm2-util to import certificates for unit tests without additional software.

alonbl:
Rebased.

@halderen: can you please review? this is super handy utility for unittest of PKCS\#11 enabled application.

jschlyter:

alonbl:
> Please rebase on develop and mark as ready when ready.

Hi @jschlyter,
Done :)
Thanks!

src/bin/util/softhsm2-util-botan.cpp
```diff
@@ -207,6 +210,53 @@ int crypto_import_key_pair
...
+	CK_CERTIFICATE_TYPE certType = CKC_X_509;
+	CK_BBOOL ckTrue = CK_TRUE, ckFalse = CK_FALSE, ckToken = CK_TRUE;
+	CK_ATTRIBUTE certTemplate[] = {
+		{ CKA_CLASS,            &certClass,       sizeof(certClass) },
+		{ CKA_CERTIFICATE_TYPE, &certType,        sizeof(certType) },
+		{ CKA_LABEL,            label,            strlen(label) },
+		{ CKA_ID,               objID,            objIDLen },
+		{ CKA_TOKEN,            &ckToken,         sizeof(ckToken) },
+		{ CKA_PRIVATE,          &ckFalse,         sizeof(ckTrue) },
+		{ CKA_VALUE,            &*blob.begin(),   (CK_ULONG)blob.size() },
```
ijsf: It doesn't seem that this function is doing error checking on the user inputs and resulting vectors.

A possible result of that is that it will be dereferencing `begin()` of an empty vector, which is not allowed (https://cplusplus.com/reference/vector/vector/begin/) and would cause undefined behaviour.

src/bin/util/softhsm2-util-ossl.cpp
```diff
@@ -216,6 +216,86 @@ int crypto_import_key_pair
...
+	X509* x509 = PEM_read_bio_X509(in, NULL, NULL, NULL);
+	BIO_free(in);
+
+	if (!x509)
+	{
+		fprintf(stderr, "ERROR: Could not read the certificate file.\n");
+		return 1;
+	}
+
+	int blobSize = i2d_X509(x509, NULL);
```
ijsf: Its advised to check if the return length (`blobSize`) is negative, in which case an error occurred (https://linux.die.net/man/3/i2d_x509).

Also for similar calls below.

ijsf (changes requested):

src/bin/util/softhsm2-util-botan.cpp
alonbl: the `Botan::X509_Certificate cert(filePath)` would fail and throw an exception.
I looked for examples where there is example of checking the `BER_encode()` result in botan and could not find any.

I changed the code so that NULL be assigned.

src/bin/util/softhsm2-util-ossl.cpp
alonbl: I rework the implementation, I hope the cleanup pattern is acceptable.

alonbl (commented):
Hello @ijsf ,

Thank you for reviewing.
I pushed a new revision, I hope it will address your concerns.

Regards,

src/bin/util/softhsm2-util.cpp
```diff
@@ -1205,6 +1248,98 @@ int importSecretKey(char* filePath, CK_SLOT_ID slotID, char* userPIN, char* labe
...
+(
+	char* filePath,
+	CK_SLOT_ID slotID,
+	char* userPIN,
+	char* label,
+	char* objectID,
+	int forceExec
+)
+{
+	char user_pin_copy[MAX_PIN_LEN+1];
```
ijsf: Initializing this array to zero `= {0};` will avoid problems later with uninitialized values.

src/bin/util/softhsm2-util-botan.cpp
ijsf: This helps, thanks for changing it so quickly.

src/bin/util/softhsm2-util-ossl.cpp
ijsf: Thanks for handling this so quickly. I know it is a bit pedantic but a check `<= 0` would improve things a bit more. This is because the manual (https://linux.die.net/man/3/i2d_x509) seems to mention:
>a negative value if an error occurs

not just `-1.`

I assume this would include zero as well.

src/bin/util/softhsm2-util.cpp
alonbl: done, although the `getPw()` will not return success without initializing it.

BTW: this is a copy of other import utilities, there are lots of issues in utilities including leaks.

ijsf: Makes sense. Might be better to do a proper cleanup in the future.

ijsf (approved):

src/bin/util/softhsm2-util-ossl.cpp
alonbl: I overlooked this, fixed.

alonbl:
Hello @jschlyter, is there anything more I can do to push this forward?

jschlyter:
I think we're good, but I'd appreciate if @bjosv could take a look as well.

src/bin/util/softhsm2-util-ossl.cpp
```diff
@@ -218,6 +218,151 @@ int crypto_import_key_pair
...
+		CK_OBJECT_CLASS certClass = CKO_CERTIFICATE;
+		CK_CERTIFICATE_TYPE certType = CKC_X_509;
+		CK_BBOOL ckTrue = CK_TRUE, ckFalse = CK_FALSE, ckToken = CK_TRUE;
+		CK_ATTRIBUTE certTemplate[] = {
+			{ CKA_CLASS,            &certClass,   sizeof(certClass) },
+			{ CKA_CERTIFICATE_TYPE, &certType,    sizeof(certType) },
+			{ CKA_LABEL,            label,        strlen(label) },
+			{ CKA_ID,               objID,        objIDLen },
+			{ CKA_TOKEN,            &ckToken,     sizeof(ckToken) },
+			{ CKA_PRIVATE,          &ckFalse,     sizeof(ckTrue) },
```
bjosv:
```suggestion
			{ CKA_PRIVATE,          &ckFalse,     sizeof(ckFalse) },
```

src/bin/util/softhsm2-util-ossl.cpp
```diff
@@ -218,6 +218,151 @@ int crypto_import_key_pair
...
+	)
+	{
+		fprintf(stderr, "ERROR: Could not convert certificate to DER.\n");
+		goto cleanup;
+	}
+
+	{
+		CK_OBJECT_CLASS certClass = CKO_CERTIFICATE;
+		CK_CERTIFICATE_TYPE certType = CKC_X_509;
+		CK_BBOOL ckTrue = CK_TRUE, ckFalse = CK_FALSE, ckToken = CK_TRUE;
```
bjosv:
```suggestion
		CK_BBOOL ckFalse = CK_FALSE, ckToken = CK_TRUE;
```

src/bin/util/softhsm2-util-ossl.cpp
bjosv: ..and then we can remove ckTrue in this function I believe.

src/bin/util/softhsm2-util-botan.cpp
```diff
@@ -209,6 +212,64 @@ int crypto_import_key_pair
...
+	CK_OBJECT_CLASS certClass = CKO_CERTIFICATE;
+	CK_CERTIFICATE_TYPE certType = CKC_X_509;
+	CK_BBOOL ckTrue = CK_TRUE, ckFalse = CK_FALSE, ckToken = CK_TRUE;
+	CK_ATTRIBUTE certTemplate[] = {
+		{ CKA_CLASS,            &certClass,          	sizeof(certClass) },
+		{ CKA_CERTIFICATE_TYPE, &certType,           	sizeof(certType) },
+		{ CKA_LABEL,            label,               	strlen(label) },
+		{ CKA_ID,               objID,               	objIDLen },
+		{ CKA_TOKEN,            &ckToken,            	sizeof(ckToken) },
+		{ CKA_PRIVATE,		&ckFalse,		sizeof(ckTrue) },
```
bjosv:
```suggestion
		{ CKA_PRIVATE,		&ckFalse,		sizeof(ckFalse) },
```

bjosv: and then we can remove `ckTrue` above.

bjosv (commented):
Looks useful, just have a few comment before a merge.

src/bin/util/softhsm2-util-botan.cpp
alonbl: correct, done!

src/bin/util/softhsm2-util-ossl.cpp
alonbl: correct, done!

src/bin/util/softhsm2-util-ossl.cpp
alonbl: correct, done!

alonbl:
@bjosv thank you for spotting this leftover!

I hope all is good now.

bjosv (approved):
LGTM!

jschlyter (approved):

jschlyter:
One more thing @alonbl, could you add a test case for the import function in a separate PR?

alonbl:
Hello @jschlyter,

I do not see any test cases for the usage of the utilities.
I may miss this, if there is infrastructure for this, please refer me and I will add this specific test case.

Thanks,
Alon

PR #614, opened 2021-03-04
Do not refresh object store before fetching object.

larssilven:
Before this commit the object store for a file token was always refreshed by reading
the file of the token every time an object of the token was fetched.
Now the HSM may be configured not to refresh when fetching an object. But the
refresh will still be done after an application gets a handle for an object.
    
The reason for this change is that the CPU time consumed by the reading may not
be negligible for some HW.

### DIFF
--- a/src/lib/object_store/SessionObject.cpp
+++ b/src/lib/object_store/SessionObject.cpp
@@ -217,7 +217,8 @@ bool SessionObject::deleteAttribute(CK_ATTRIBUTE_TYPE type)
 }
 
 // The validity state of the object
-bool SessionObject::isValid()
+// the doRefresh parameter has no meaning for this implementation since noting is stored on disk.
+bool SessionObject::isValid(const bool doRefresh __attribute__((unused)))
 {
     return valid;
 }
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -609,6 +609,8 @@ CK_RV SoftHSM::C_Initialize(CK_VOID_PTR pInitArgs)
 	// Load the handle manager
 	handleManager = new HandleManager();
 
+	doRefresh = Configuration::i()->getBool("objectstore.readrefresh", true);
+
 	// Set the state to initialised
 	isInitialised = true;
 
@@ -1605,7 +1607,7 @@ CK_RV SoftHSM::C_CopyObject(CK_SESSION_HANDLE hSession, CK_OBJECT_HANDLE hObject
 
 	// Check the object handle.
 	OSObject *object = (OSObject *)handleManager->getObject(hObject);
-	if (object == NULL_PTR || !object->isValid()) return CKR_OBJECT_HANDLE_INVALID;
+	if (object == NULL_PTR || !object->isValid(doRefresh)) return CKR_OBJECT_HANDLE_INVALID;
 
 	CK_BBOOL wasOnToken = object->getBooleanValue(CKA_TOKEN, false);
 	CK_BBOOL wasPrivate = object->getBooleanValue(CKA_PRIVATE, true);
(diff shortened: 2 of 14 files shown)


rijswijk:
Thanks for your PR; the reason it is implemented the way it currently is is to cope with multiple independent processes concurrently manipulating the same token through the SoftHSM library. If you disable the check, then a process will fail to pick up any changes to the token made by an external process. I understand that for performance reasons this is inconvenient, but perhaps it would be better, if you want to be able to disable it, to make this a configurable feature?

I'm wondering, though, if that should then be run-time configurable or compile-time (since you might not want two processes using SoftHSM with conflicting configurations). I'd like to ask @halderen for an opinion on this too.

halderen:
It is really needed that multiple concurrent processes access the same token at the same time for many use cases.  I'd first like to know the performance impact of such a thing. SoftHSMv2 isn't really that fast anyway (for security reasons) so if the effect isn't good for me.  Not having the lock (so this PR) means bringing back the old SoftHSMv1 functionality and the way SoftHSMv2 sometimes operate due to a bug.  This took us with OpenDNSSEC 2 1/2 years to find, so I'd rather not burn application programmers with this.  In any case, this should not be a compile time option, as this would mean package maintainers may select this, and then everyone may be stuck with this.  It should be a run-time, per slot/token option (but that might not be possible) option that is not default on.

larssilven:
I'm glad that you are willing to accept my proposal with the modification that the 'refresh disabling' should be optional and disabled by default.

Could we add a key called objectstore.readrefresh to the softhsm2.conf . If this key has the value 'false' then no refreshing is done before fetching an object from the store, if something else or not define then the refresh is done (as before).

I have just tested this with the file mode. Maybe nothing should be done about the DB mode (I removed a mutex lock)?

I believe it should be possible in 'file mode' for several processes to use the same objects in a store with 'objectstore.readrefresh=false'. There will just be a problem if one of the processes use C_DestroyObject or C_SetAttribute;
then the other processes will continue to used the object as it where before.

larssilven:
New version ready for review.

larssilven:
### Benchmarks

storage type | objectstore.readrefresh | objectstore.backend | signes/s
-----------------|---------------------------------|----------------------------|-------------
9p | true | file | 386
9p | true | db | 1275
9b | false | file | 2393
9b | false | db | 1263
qcow2 | true | file | 2534
qcow2 | true | db | 2170
qcow2 | false | file | 2379
qcow2 | false | db | 2163

jschlyter:

larssilven:
DES tests are failing as you can see.
I have disabled DES in the build for my project. If I remember correctly it had to do with the openssl on debian.
My other PR #643 is not failing on github but fails on my own machine when I build with an unmodified config.
I will see what I could do. Maybe I could compare with the other ciphers that is working on Debian.

It is great that this project is up and running again. I need it in my project and have not found anything similar I could use.
Thank you!!!!

larssilven:
Now all tests but windows are OK. But I don't think these problem has anything to do with this.

src/lib/object_store/SessionObject.cpp
```diff
@@ -217,7 +217,8 @@ bool SessionObject::deleteAttribute(CK_ATTRIBUTE_TYPE type)
 }
 
 // The validity state of the object
-bool SessionObject::isValid()
+// the doRefresh parameter has no meaning for this implementation since noting is stored on disk.
+bool SessionObject::isValid(const bool doRefresh __attribute__((unused)))
```
bjosv: The build fails for windows with
`SessionObject.cpp(221,50): error C3646: '__attribute__': unknown override specifier`

You might need to replace `const bool doRefresh __attribute__((unused))` with something like `const bool /* doRefresh */`, or adding `(void)doRefresh;` in the function body.

larssilven:
how stupid of me. I did just assume the the windows build was not ready yet.
Thanks @bjosv

src/lib/object_store/SessionObject.cpp
larssilven: how stupid of me. I did just assume the the windows build was not ready yet.
Thanks @bjosv

PR #616, opened 2021-03-11
Support GOST 2012 256 with botan

gecheim:
In these commits the following changes:

- Fix build with CMake
- Added support CMake build with static botan library
- Added support work with GOST R 34.10 2012 256 keys with botan backend.


rijswijk:
Thank you for the PR; I gave it a quick review, I saw some changes in the OpenSSL code base as well, the PR suggests it is only for Botan, do I see correctly that these are only there to make the code path for `getGOSTType(..)` complete? If so, is this the right approach since it seems GOST 2012 is not actually supported with OpenSSL?

I will also ask @halderen for a more detailed review.

gecheim:
Yes, changes in OpenSSL code base support only `getGOSTType(..)` otherwise the project would not build with the OpenSSL backend.

OpenSSL support GOST 2012 with engine.

These changes were made for a task where you only need to work with the botan backend. So the other option was not supported in this pull request.

atribolt:
We have the same problem. We need to work with keys like Gost2012-256 using botan. Could you apply this patch to support working with botan? But for Gost 2012 with openssl, create a separate issue.

jschlyter:

Issue #623, opened 2021-03-30
Use authenticated encryption in SecureDataManager


rgerganov:
Right now the SecureDataManager is using AES in CBC mode to protect the master key derived from the PIN and sensitive object attributes. The problem with this approach is that there is no way to guarantee the integrity of the object store. If an attacker gets access to the filesystem, they can modify the object files leaving this undetected by the SoftHSM. In this situation users will get incorrect results when using SoftHSM instead of error saying that the store has been tampered.

This problem can be easily solved by replacing AES-CBC with some of the authenticated modes of AES such as AES-GCM. SoftHSM already supports AES-GCM, so the code changes in SecureDataManager.cpp would be minimal. However, there will be some work for backward compatibility with old tokens.

I wonder what others think about this. I can start working on a PR if we reach some agreement.

rijswijk:
Sounds like a worthwhile enhancement, if you would be willing to submit a PR that would be much appreciated. I do think this is a major change that requires at least a bump in minor version number, maybe a configurable option and a conversion tool for existing tokens, though. @halderen do you have thoughts on this?

rgerganov:
I put some notes while working on this here: https://xakcop.com/post/softhsmv2/
Could be helpful to other people interested into how SoftHSM works.

Aearsis:
If I'm not missing anything, this still does not protect fully from tampering. As the store isn't encrypted as a whole, an attacker can delete objects or object attributes, effectively turning them into default values. This is possible for both file and database backends.

Also, this PR doesn't protect against replaying old values (backups), but that's inevitable in case of SoftHSM - just a reminder.

Otherwise, the change itself looks good to me and is still better than no integrity at all :+1:

PR #627, opened 2021-04-13
Replace AES-CBC with AES-GCM in SecureDataManager

rgerganov:
Use AES-GCM for encrypting the master key (derived from the PIN) and
get rid of the magic bytes.
Use AES-GCM for encrypting object attributes and thus guarantee the
integrity of the object files (i.e. detect if the object store has been
tampered).

Issue: #623

### DIFF
--- a/src/lib/data_mgr/SecureDataManager.cpp
+++ b/src/lib/data_mgr/SecureDataManager.cpp
@@ -65,9 +65,6 @@ void SecureDataManager::initObject()
 	// Set the initial login state
 	soLoggedIn = userLoggedIn = false;
 
-	// Set the magic
-	magic = ByteString("524A52"); // RJR
-
 	// Get a mutex
 	dataMgrMutex = MutexFactory::i()->getMutex();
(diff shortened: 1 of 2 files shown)


rijswijk:
Thanks for the contribution! This indeed looks like a minimal intervention, although I think it should be a configurable option rather than an immediate change, and it is likely also good to detect what type of encryption an existing token uses, since users that upgrade SoftHSM versions will want to keep using their existing tokens. I realise this may be quite a bit of work, if you don't have time to do that, I suggest we put the PR on hold for now.

rgerganov:
Agree, this needs more work for backward compatibility with existing tokens. I just wanted to push my changes so other people can review and provide feedback. I hope I will be able to find time to complete this in the near future. Also if anyone is willing to collaborate they are free to jump in.

rijswijk:
OK, thanks, I will tag it as "on hold". I appreciate your patience and contribution.

jschlyter:

PR #629, opened 2021-04-19
fix check minimal Botan version

agrandville:
compilation with Botan crypto backend on Ubuntu 18 fails.

way to reproduce:
```
$ sudo apt install libbotan-2-dev
$ sudo apt list --installed | grep botan
libbotan-2-dev/bionic,now 2.4.0-5ubuntu1 amd64 [installed]
$ ./autogen.sh
$ ./configure --with-crypto-backend=botan 
$ make

 ....lot of output....

BotanUtil.cpp:172:13: error: ‘class Botan::BER_Object’ has no member named ‘is_a’
  if (object.is_a(Botan::PRINTABLE_STRING, Botan::ASN1_Tag(0)))
             ^~~~
```

BER_Object is defined in include/botan-2/botan/asn1_obj.h
Member function 'is_a' was added in [Botan version 2.5](https://github.com/randombit/botan/blob/2.5.0/src/lib/asn1/asn1_obj.h)

### DIFF
--- a/m4/acx_crypto_backend.m4
+++ b/m4/acx_crypto_backend.m4
@@ -119,7 +119,7 @@ AC_DEFUN([ACX_CRYPTO_BACKEND],[
 	elif test "x${crypto_backend}" = "xbotan"; then
 		AC_MSG_RESULT(Botan)
 
-		ACX_BOTAN(2,0,0)
+		ACX_BOTAN(2,5,0)
 
 		CRYPTO_INCLUDES=$BOTAN_CFLAGS
 		CRYPTO_LIBS=$BOTAN_LIBS


rijswijk:
Thanks for spotting this, looks like this was introduced with #526 and fell through the cracks. I wonder, though, if there is a different solution possible that doesn't use `is_a(..)`, since bumping the Botan version might break builds on older platforms.

Question for @Jakuje who submitted #526: do you perhaps see an alternative to using `is_a(..)`?

Jakuje:
Probably, there will be a way, but this was literally my first code I wrote with botan so anyone should be eligible to figure out some alternative. If nobody will show up, I will have a look, but I can not promise when this will be.

halderen:
I should re-institute running the tests on Jenkins on all platforms to catch these items, the testing platforms atm aren't cutting it.
Nearly all distributions are on Botan version > 2.5.0 according to https://github.com/randombit/botan/wiki/Distros
Ubuntu 18 is still into LTS until mid 2022, Ubuntu 18 is on a too old version of botan, but also softhsm version lacks very behind.  Hence demanding a higher version of ubuntu doesn't hurt, as people need to manually build softhsm2 anyway, and can then do the same to Botan.  Ubuntu 20 is on >= 2.5.0 so fine.

PR #632, opened 2021-05-13
Replace CKR_ARGUMENTS_BAD by a better error code when appropriate

selvanair:
Cases where the error is due to a limitation in the
implementation (e.g., digest type in OAEP), return
CKM_MECHANISM_PARAMETER_INVALID instead of CKR_ARGUMENTS_BAD

Where the parameter is evidently wrong (like NULL for a
required parameter), CKR_ARGUMENTS_BAD is retained.

Signed-off-by: Selva Nair <selva.nair@gmail.com>


selvanair:
PS: I'm not sure of how to treat the `source == CKZ_DATA_SPECIFIED`
requirement imposed for OAEP in some code paths. Is this required 
to be set even when pSourceData is unused (NULL)? 
Also, for some reason its not checked during Decrypt() --  only 
Encrypt/Wrap/UnWrap checks it.

jschlyter:

PR #639, opened 2021-07-15
According to the PKCS11 specification, modify the return value type o…

c20liu:
According to the PKCS11 specification(v2.4 or v3.0), some key-related operations need return CKR_KEY_HANDLE_INVALID


jschlyter:

PR #641, opened 2021-07-29
add wrapping and unwrapping capability to CKM_AES_GCM mechanism

keldonin:
This PR adds wrapping and unwrapping capability to the `CKM_AES_GCM` mechanism.
It comes along with new cases to the test suite.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -1162,7 +1162,7 @@ CK_RV SoftHSM::C_GetMechanismInfo(CK_SLOT_ID slotID, CK_MECHANISM_TYPE type, CK_
 		case CKM_AES_GCM:
 			pInfo->ulMinKeySize = 16;
 			pInfo->ulMaxKeySize = 32;
-			pInfo->flags |= CKF_ENCRYPT | CKF_DECRYPT;
+			pInfo->flags |= CKF_ENCRYPT | CKF_DECRYPT | CKF_WRAP | CKF_UNWRAP;
 			break;
 		case CKM_AES_KEY_WRAP:
 			pInfo->ulMinKeySize = 16;
@@ -2197,7 +2197,7 @@ CK_RV SoftHSM::SymEncryptInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMech
 	size_t bb = 8;
 	size_t counterBits = 0;
 	ByteString aad;
-	size_t tagBytes = 0;
+	size_t tagBits = 0;
 	switch(pMechanism->mechanism) {
 #ifndef WITH_FIPS
 		case CKM_DES_ECB:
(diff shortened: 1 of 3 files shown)


keldonin:
The build seems to fail with Appveyor. However, when looking closer to the logs:
- all builds are on MSVC/Windows (no build on linux/BSD?)
- the builds fail as `syslog.h` cannot be found (and is probably not expected on this platform). 

Is something broken in the source code, that prevents building on Windows?

```
"C:\projects\softhsmv2\softhsm2.sln" (Build target) (1) ->
"C:\projects\softhsmv2\ALL_BUILD.vcxproj.metaproj" (default target) (2) ->
"C:\projects\softhsmv2\src\lib\softhsm2.vcxproj.metaproj" (default target) (4) ->
"C:\projects\softhsmv2\src\lib\common\softhsm_common.vcxproj.metaproj" (default target) (5) ->
"C:\projects\softhsmv2\src\lib\common\softhsm_common.vcxproj" (default target) (6) ->
(ClCompile target) -> 
  c:\projects\softhsmv2\src\lib\common\log.h(40): fatal error C1083: Cannot open include file: 'syslog.h': No such file or directory [C:\projects\softhsmv2\src\lib\common\softhsm_common.vcxproj]
```

saper:
@keldonin I found another issue with AES GCM and have a fix in #664. Since the fix would have need to be applied also to the wrapping code, maybe we should think about extracting those AES parameter checks into a common function. I have also refactored the test https://github.com/opendnssec/SoftHSMv2/pull/664/commits/d1553412e61bdf73281f0a8d36f5df2406bfaf79 to be able to test different sets of parameters.

src/lib/SoftHSM.cpp
```diff
@@ -6298,9 +6302,51 @@ CK_RV SoftHSM::WrapKeySym
...
 
+		case CKM_AES_GCM:
+			if (pMechanism->pParameter == NULL_PTR ||
+			    pMechanism->ulParameterLen != sizeof(CK_GCM_PARAMS))
+			{
+				DEBUG_MSG("GCM mode requires parameters");
+				return CKR_ARGUMENTS_BAD;
+			}
+			iv.resize(CK_GCM_PARAMS_PTR(pMechanism->pParameter)->ulIvLen);
+			memcpy(&iv[0], CK_GCM_PARAMS_PTR(pMechanism->pParameter)->pIv, CK_GCM_PARAMS_PTR(pMechanism->pParameter)->ulIvLen);
```
rijswijk, line 6319: I think this requires the same fix as in #664

src/lib/SoftHSM.cpp
```diff
@@ -6298,9 +6302,51 @@ CK_RV SoftHSM::WrapKeySym
...
+			aad.resize(CK_GCM_PARAMS_PTR(pMechanism->pParameter)->ulAADLen);
+			memcpy(&aad[0], CK_GCM_PARAMS_PTR(pMechanism->pParameter)->pAAD, CK_GCM_PARAMS_PTR(pMechanism->pParameter)->ulAADLen);
```
rijswijk, line 6321: I think this requires the same fix as in #664

jschlyter:

PR #643, opened 2021-07-30
No memory allocation for each C_DecryptUpdate.

larssilven:
Before this change the allocated memory was growing after each
C_DecryptUpdate.
If a large amount of data is decrypted then the memory runs out ...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decryption handling to ensure encrypted data is only buffered in GCM mode, enhancing reliability for users in certain encryption scenarios.


larssilven:
The 'currentAEADBuffer' that is growing for each update is just used for the GCM decryption. So my easy fix is just to not update this buffer when another cipher mode but GCM is used.
This is of course not fixing the problem for the GCM that also has to be fixed.
Maybe it would be better if each cipher mode should have its own class instead of using this 'mode' variable.

PR #648, opened 2021-09-06
Fix C_GetMechanismInfo to fail on non-allowed mechanisms

Jakuje:


src/lib/SoftHSM.h
```diff
@@ -489,7 +489,7 @@ class SoftHSM
 
 	CK_RV MechParamCheckRSAPKCSOAEP(CK_MECHANISM_PTR pMechanism);
 
-	bool isMechanismPermitted(OSObject* key, CK_MECHANISM_PTR pMechanism);
+	bool isMechanismPermitted(OSObject* key, CK_MECHANISM_TYPE pMechanism);
```
ueno: nit: `pMechanism` -> `mechanism` now that it's not a pointer?

src/lib/SoftHSM.cpp
```diff
@@ -12830,22 +12831,26 @@ CK_RV SoftHSM::MechParamCheckRSAPKCSOAEP(CK_MECHANISM_PTR pMechanism)
...
 
-	OSAttribute attribute = key->getAttribute(CKA_ALLOWED_MECHANISMS);
-	std::set<CK_MECHANISM_TYPE> allowed = attribute.getMechanismTypeSetValue();
+	/* If we have object, consult also its allowed mechanisms */
+	if (key) {
+		OSAttribute attribute = key->getAttribute(CKA_ALLOWED_MECHANISMS);
+		std::set<CK_MECHANISM_TYPE> allowed = attribute.getMechanismTypeSetValue();
 
-	if (allowed.empty()) {
+		if (allowed.empty()) {
```
ueno, line 13272: nit: Is this special case needed?

src/lib/SoftHSM.h
Jakuje: Thanks. Will fix.

src/lib/SoftHSM.cpp
Jakuje: Yes. The empty "allowlist" means that everything that is built-in is allowed. The followed check would result in "false" return value for all mechanisms if no allow list would be specified. But this deserves a comment, which I will add here.

Jakuje:
> @Jakuje Could you please rebase this branch on the latest develop and request a review again ? Thanks !

I can not seem to be able to request review, but the PR is rebased and looks like still working after 4 years anniversary! So ready for a review!

Issue #628, opened 2021-04-16
Unwrapped AES keys have VALUE_LEN set to 0


ceridwen:
I'm using the Python PKCS (https://github.com/danni/python-pkcs11) library to unwrap some AES keys wrapped by another HSM. While the unwrapping process seems to work (no error from the library or in SoftHSM's debug logs), CKA_VALUE_LEN for these (256 bit) keys seems to be 0: 

```
            example_unwrapping_test = example_wrapping_softhsm.unwrap_key(pkcs11.ObjectClass.SECRET_KEY, pkcs11.KeyType.AES, example_blob, mechanism=pkcs11.Mechanism.RSA_PKCS_OAEP, store=True, label='example-unwrapping-test', template={
            pkcs11.Attribute.SENSITIVE: False,
            pkcs11.Attribute.EXTRACTABLE: True,
        })
        except Exception as error:
            print(error)
            print(vars(error))
            raise
        print(example_unwrapping_test)
        print(len(example_unwrapping_test[pkcs11.Attribute.VALUE]))
        print(example_unwrapping_test[pkcs11.Attribute.VALUE_LEN])
```

```
<SecretKey label='example-unwrapping-test' id='' 0-bit AES>
32
0
```

The actual bytes of the test key are the right length (32) and do have the same value as the test key before it was wrapped.

The standard implies that, at least for RSA_OAEP, CKA_VALUE_LEN for these keys should come from the mechanism: "[T]he mechanism contributes only the CKA_CLASS and CKA_VALUE (and CKA_VALUE_LEN, if the key has it) attributes to the recovered key during unwrapping."  I see this when unwrapping an AES key wrapped with CKM_AES_KEY_WRAP_PAD too.

rijswijk:
Have you explicitly included the `CKA_VALUE_LEN` attribute in the (optional) unwrapping template? If I look at the code, that should do the trick.

ceridwen:
It looks like when I try to set that field, SoftHSM is telling me that the attribute is read-only:

```
Traceback (most recent call last):
  File "offline_example.py", line 542, in <module>
    'example-wrapped',
  File "offline_example.py", line 307, in unwrap_key
    pkcs11.Attribute.VALUE_LEN: 256,
  File "pkcs11/_pkcs11.pyx", line 1276, in pkcs11._pkcs11.UnwrapMixin.unwrap_key
  File "pkcs11/_errors.pyx", line 88, in pkcs11._pkcs11.assertRV
[...]

PR #650, opened 2021-09-23
fix for issue #628 - CKA_VALUE_LEN is synthetized on every secret key

keldonin:
This little patch (including test code) will synthetize a `CKA_VALUE_LEN` attribute for all secret keys, whenever `C_UnwrapKey()` is invoked. This should address issue #628.


jschlyter:

Issue #405, opened 2018-07-05
Support CKM_AES_CBC_PAD for the wrapping functions


lingxiankong:
Could we implement CKM_AES_CBC_PAD for wrapping in order to support SoftHSM in OpenStack Barbican. Here is Barbican related code: https://github.com/openstack/barbican/blob/5dea5cec130b59ecfb8d46435cd7eb3212894b4c/barbican/plugin/crypto/pkcs11.py#L496

Or is there some solution could work around this?

bellgrim:
As noted in #229, this mechanism is not implemented. But there are other wrapping mechanisms available. Keeping this issue open as a feature request.

lingxiankong:
@bellgrim hi thanks for the reply. I'm not familiar with PKCS programing, do you mean I just need to replace CKM_AES_CBC_PAD with either CKM_AES_KEY_WRAP, CKM_AES_KEY_WRAP_PAD, CKM_RSA_PKCS, or CKM_RSA_PKCS_OAEP in the barbican code?

bellgrim:
Yes, there are other wrapping mechanisms but it is probably not as easy as just swapping the name of the mechanism when doing the operation. I have not implemented them so I do not know the details of each method. You can find more information about the mechanisms in PKCS#11 v2.40. It then points to e.g. RFC5649 for CKM_AES_KEY_WRAP_PAD.

zinoadidi:
Hello
Is it possible to get a link to sample implementation / how to switch to CKM_AES_KEY_WRAP_PAD in Java?

rijswijk:
We generally don't provide programming examples, and are not familiar enough with Java to help you with that, we suggest you ask this question on StackExchange or a similar forum.

pedrolamarao:
I would like to second this request. Using `CKM_DES3_CBC` or `CKM_AES_CBC` to encrypt/unwrap private keys is established practice. This is blocking integration of our applications with SoftHSM2.

rijswijk:
There are security reasons to use other mechanisms, we cannot promise that these older mechanisms will be implemented anytime soon, but we would welcome a pull request if that is an option.

rajivmucheli:
Hi, is this issue resolved ?

Was anyone able to switch the wrapping mechanism of Openstack Barbican from CKM_AES_CBC_PAD to CKM_AES_KWP ?

PR #657, opened 2021-10-29
Add support for key wrap based on 3DES

threez:
Addresses the issue in https://github.com/opendnssec/SoftHSMv2/issues/405


jschlyter:

PR #665, opened 2022-01-11
Support encryption for END_DATE and START_DATE attributes in private keys

### LINKED ISSUE #655: Attributes START_DATE and END_DATE are not ciphered in a private key
Hello,

While developing a product for which we have to support multiple HSM,  we encountered a problem in SoftHSM with the START_DATE and END_DATE attrtibutes for private keys.
It is a case that is supported by the spec but from what I understand, SoftHSM expects that START_DATE and END_DATE are only set for public keys.
When set on private keys and reloading the private key and reading the _DATE attributes, it tries to decrypt them but fails in IV reading since they are not encrypted to begin with.

antoinelochet:
Addresses the issue in #655 

## Summary by CodeRabbit

* **Bug Fixes**
  * Start/end date attributes are now stored reliably, respect privacy (encrypted when marked private), and are handled correctly during object creation.

* **Tests**
  * Added RSA public/private key tests validating start/end date attribute behavior on object creation.

* **Chores**
  * Improved debug/error logging for template processing and data decryption; updated CI checkout/tooling versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     name: Linux with Botan
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - name: Prepare
         run: |
           sudo apt-get update -qq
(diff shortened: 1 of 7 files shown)


jschlyter:

antoinelochet:
Hello @jschlyter 
Done :)

antoinelochet:
As requested by @jschlyter, this PR has been rebased to the latest develop

PR #671, opened 2022-03-14
Check RSA-OAEP mechanims when decrypting

Jakuje:
The same check is in all the other methods handling the RSA-OAEP
encryption, wrapping and unwrapping, but for some reason, it was missing
in the decryption operation.

Related: https://github.com/OpenSC/libp11/pull/440


jschlyter:

Jakuje:
> Please rebase on develop and mark as ready when ready.

done

bjosv (approved):
Good finding, we will check that `params->source == CKZ_DATA_SPECIFIED` now.
I wonder if we could improve our tests in the near future to cover this, any idea?

Jakuje:
We discussed the `CKZ_DATA_SPECIFIED` issue and the specs is not exactly clear if it should be always 1 ( CKZ_DATA_SPECIFIED) or it could be also 0 in cases there are no additional data. But this is mostly on a side and should not affect this PR.

Indeed, more tests could help for all the operations.

PR #677, opened 2022-05-30
Fix memory leak in SoftHSM::UnwrapKeySym.

dcoombs:
When using C_UnwrapKey with CKM_xxx_CBC_PAD, the unwrapping key and
cipher objects were not being recycled on success.


dcoombs:
(The appveyor failures seem to be about downloading openssl, not my fault :) )

PR #681, opened 2022-08-16
Fix severe peformance issue accessing keys

### LINKED ISSUE #680: Lookups of specific keys are too slow
The OpenDNSSEC `ods-hsmutil list` command is extremely slow.

First it does a lookup of all private keys using `C_FindObjects*` which takes 2-3 seconds because there are over 300 of them.

Then it does a lookup of the public key corresponding to each private key using `C_FindObjects*` which takes 2-3 seconds **per key** because it goes through all of the keys again.

There doesn't appear to be a better PKCS#11 API for doing this so HSMs must be expected to have faster lookup processes, at least when a specific key ID is provided.

The list command can be improved but key access shouldn't take 2+ seconds.

nomis:
The "true" in the call to Generation::create() in OSToken::OSToken() is
used as the umask when it's supposed to be the isToken value (#566).

Remove the default value from isToken because it's dangerous and there are
only two callers. Explicitly pass "true" and "false" for isToken.

Failing to consider this a token generation file means that the value is
never refreshed for read-only operations. All objects are reloaded from
disk every time one of them is refreshed. List operations take a long time
because all of the objects are re-read for each object.

Fixes #680.

### DIFF
--- a/src/lib/object_store/Generation.cpp
+++ b/src/lib/object_store/Generation.cpp
@@ -35,7 +35,7 @@
 #include "Generation.h"
 
 // Factory
-Generation* Generation::create(const std::string path, int umask, bool isToken /* = false */)
+Generation* Generation::create(const std::string path, int umask, bool isToken)
 {
 	Generation* gen = new Generation(path, umask, isToken);
 	if ((gen != NULL) && isToken && (gen->genMutex == NULL))
--- a/src/lib/object_store/Generation.h
+++ b/src/lib/object_store/Generation.h
@@ -42,7 +42,7 @@ class Generation
 {
 public:
 	// Factory
-	static Generation* create(const std::string inPath, int inUmask, bool inIsToken = false);
+	static Generation* create(const std::string inPath, int inUmask, bool inIsToken);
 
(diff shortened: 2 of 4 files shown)


nomis:
`ods-hsmutil list` now takes 2.7s instead of 13+ minutes for 309 keys...

sarlott:
@halderen it looks like you're nearing the point of making a release for other changes, could you merge this soon?

nomis:
@halderen it looks like you're nearing the point of making a release for other changes, could you merge this soon?

ijsf (approved):
Its pretty obvious this is an argument handling mistake in the code. This PR fixes that and makes the function call more explicit.

bjosv (approved):
Looks good!
A fixup for #566.
Maybe there is clang-tidy checker that could have helped us catching this, [implicit-bool-conversion](https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html) might be to broad but maybe there is something else in that toolbox.

jschlyter (approved):

Issue #620, opened 2021-03-19
Support for CKM_ECDSA with hashing


johughes99:
I have been undertaking a wide range of performance tests - including for signing/verification.

Whilst I understand that support for SHA3 is an enhancement request I was slightly surprised that mechanisms such as CKM_ECDSA_SHA256 and CKM_ECDSA_SHA512 are not supported

Any reasons why only CKM_ECDSA is supported as regards a  EC signing mechanism?

Is it because there is no demand - and/or that typical applications compute the hash prior to calling C_SignInit with the CKM_ECDSA mechanism?

jca-klk:
I was also surprised by the lack of support of CKM_ECDSA_SHA256.
I gave #683 a spin, and it works for my small tests scenario.

Any chance this gets merged ?

cryptoquick:
I also ran into this just now. I installed the binary from the Arch package repo, so now either I'll need to change my code or I'm going to have to figure out how to build this...

PR #683, opened 2022-08-21
Support for ECDSA with hashing

MatthiasValvekens:
Hi, I had a go at #620, since it'd be useful for one of my projects, in which I use SoftHSM as a testing backend for PKCS#11 code.

The PR is very much a work in progress: it does what I want with both the OpenSSL backend and the Botan one, but I'd like to get an extra pair of eyes before I start adding tests.

Thanks!


antoinelochet:
Hello @MatthiasValvekens,
I need your work on this PR for my project too.
But I am missing all the verification logic and the mechanisms info.
So I would like to contribute to your PR.
Could you tell me if you are ok with my proposal for contribution ?

MatthiasValvekens:
Sure, I sorta gave up on it at the time since I didn't get any response (and didn't want to invest too much time only for the PR to never get merged in the end), but absolutely feel free to build upon what I did!

PR #694, opened 2023-01-30
avoid unnecessary check for sqlite3 binary

jluebbe:
Only the library is used, not the sqlite3 binary. Drop this check to simplify cross-compilation (as no native sqlite3 binary is needed).


halderen:
Agreed, I was wondering whether the binary might have been in use for testing purposes, but it doesn't look like that (not that that would have been blocking).

PR #710, opened 2023-05-08
Limit number of sessions

istepic:
When SoftHSMv2 is used for testing PKCS#11 client code before hardware is available, maximum session count on a token needs to be set in order to detects bugs in client code where sessions are leaked and remain open. This can cause problems since most hardware implementations have a limit on opened sessions count.

This change sets a limit to general session count.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -158,6 +158,11 @@ AC_DEFINE_UNQUOTED(
 	[Minimum PIN length]
 )
 AC_DEFINE_UNQUOTED(
+	[MAX_SESSION_COUNT],
+	[16],
+	[Maximum session count]
+)
+AC_DEFINE_UNQUOTED(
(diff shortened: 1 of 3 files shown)


sarlott:
Shouldn't this default to the original value or be runtime configurable? 

Those of us using this entirely in software don't want arbitrary low session limits.

nomis:
Shouldn't this default to the original value or be runtime configurable?

Those of us using this entirely in software don't want arbitrary low session limits.

bjosv (commented):
Additionally, I believe we also need to provide new defines when building with CMake (see config.h.in.cmake and CMakeLists.txt).

jschlyter:

bukka:
I agree that this way too low default and it should be runtime configurable (ideally through the config).

PR #724, opened 2023-09-01
Fixed use of ECB with Botan

antoinelochet:
Here is my proposed fix for issue 723

### DIFF
--- a/src/lib/data_mgr/SecureDataManager.cpp
+++ b/src/lib/data_mgr/SecureDataManager.cpp
@@ -51,487 +51,496 @@
 // Initialise the object; called by all constructors
 void SecureDataManager::initObject()
 {
-	// Get an RNG instance
-	rng = CryptoFactory::i()->getRNG();
+    // Get an RNG instance
+    rng = CryptoFactory::i()->getRNG();
 
-	// Get an AES implementation
-	aes = CryptoFactory::i()->getSymmetricAlgorithm(SymAlgo::AES);
+    // Get an AES implementation
+    aes = CryptoFactory::i()->getSymmetricAlgorithm(SymAlgo::AES);
 
-	// Initialise masking data
-	mask = new ByteString();
+    // Initialise masking data
+    mask = new ByteString();
 
-	rng->generateRandom(*mask, 32);
+    rng->generateRandom(*mask, 32);
 
-	// Set the initial login state
-	soLoggedIn = userLoggedIn = false;
+    // Set the initial login state
+    soLoggedIn = userLoggedIn = false;
 
-	// Set the magic
-	magic = ByteString("524A52"); // RJR
+    // Set the magic
+    magic = ByteString("524A52"); // RJR
 
-	// Get a mutex
-	dataMgrMutex = MutexFactory::i()->getMutex();
+    // Get a mutex
+    dataMgrMutex = MutexFactory::i()->getMutex();
 }
 
 // Constructs a new SecureDataManager for a blank token; actual
 // initialisation is done by setting the SO PIN
 SecureDataManager::SecureDataManager()
 {
-	initObject();
+    initObject();
 }
 
 // Constructs a SecureDataManager using the specified key blob
 SecureDataManager::SecureDataManager(const ByteString& soPINBlob, const ByteString& userPINBlob)
 {
-	initObject();
+    initObject();
 
-	// De-serialise the key blob
-	soEncryptedKey = soPINBlob;
-	userEncryptedKey = userPINBlob;
+    // De-serialise the key blob
+    soEncryptedKey = soPINBlob;
+    userEncryptedKey = userPINBlob;
 }
 
 // Destructor
 SecureDataManager::~SecureDataManager()
 {
-	// Recycle the AES instance
-	CryptoFactory::i()->recycleSymmetricAlgorithm(aes);
+    // Recycle the AES instance
+    CryptoFactory::i()->recycleSymmetricAlgorithm(aes);
 
-	// Clean up the mask
-	delete mask;
+    // Clean up the mask
+    delete mask;
 
-	MutexFactory::i()->recycleMutex(dataMgrMutex);
+    MutexFactory::i()->recycleMutex(dataMgrMutex);
 }
 
 // Generic function for creating an encrypted version of the key from the specified passphrase
 bool SecureDataManager::pbeEncryptKey(const ByteString& passphrase, ByteString& encryptedKey)
 {
-	// Generate salt
-	ByteString salt;
+    // Generate salt
+    ByteString salt;
 
-	if (!rng->generateRandom(salt, 8)) return false;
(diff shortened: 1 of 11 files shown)


src/lib/data_mgr/SecureDataManager.cpp
```diff
@@ -397,6 +397,7 @@ void SecureDataManager::logout()
 // Decrypt the supplied data
 bool SecureDataManager::decrypt(const ByteString& encrypted, ByteString& plaintext)
 {
+    DEBUG_MSG("encrypted: %s", encrypted.hex_str().c_str());
```
bjosv, line 400: This file uses tabs, but here we have 4 spaces for indentation

bjosv (commented):
Seems reasonable.
Are all the extra debug printouts needed? I seems that existing printouts are quite sparse with information, maybe for security reasons(?).

As an addition I suggest that we also disable 2 testcases when built WITH_BOTAN.
They will always fail when using Botan as I understand it. WDYT?

```
1) test: AESTests::testECB (F) line: 518 AESTests.cpp
assertion failed
- Expression: aes->encryptInit(&aesKey128, SymMode::ECB, IV)


2) test: DESTests::testECB (F) line: 537 DESTests.cpp
assertion failed
- Expression: des->encryptInit(&desKey56, SymMode::ECB, IV)
```

src/lib/data_mgr/SecureDataManager.cpp
antoinelochet: I have deleted the debug messages and reindented all the `SecureDataManager.cpp` file

antoinelochet:
I think that you are right about the tests.

I have added the debug logs because they are often useful. I agree that some security reasons, it may seem bad, but:
- using SoftHSM in production is madness
- using SoftHSM in production WITH debug logs is beyond madness
So I thought this was tolerable. But per your request, I deleted all the offending lines.

src/lib/crypto/test/AESTests.cpp
```diff
@@ -328,6 +328,7 @@ void AESTests::testCBC()
 	}
 }
 
+#ifndef WITH_BOTAN
```
bjosv: I believe this will give an undefined reference to this function while linking.
Maybe its better to just remove the registration of the testcase in the header file.

```
#ifndef WITH_BOTAN
        CPPUNIT_TEST(testECB);
#endif
```

bjosv: Same for the `testECB()` testcase in:
src/lib/crypto/test/DESTests.h

bjosv (commented):
Good comments. Using debug logs in production would be "beyond madness" :)

src/lib/crypto/test/AESTests.cpp
antoinelochet: I followed your suggestion and compilation/linking is ok now.
The tests fail on the src/lib/test though but I am pretty sure that it was the case before my changes.

bjosv: Great! Looks good.

anpa8480:
if you revert BotanSymmetricAlgorithm.cpp 8fd89ec33edfe4a7b510c471101f78b52a0c1294 and uncomment all tests are passing

antoinelochet:
> if you revert BotanSymmetricAlgorithm.cpp [8fd89ec](https://github.com/opendnssec/SoftHSMv2/commit/8fd89ec33edfe4a7b510c471101f78b52a0c1294) and uncomment if (mode == SymMode::ECB) all tests are passing.
> 
> Tested with Botan 2.19.3 image

That's because ECB is not taken into account during compilation phase. It does not fix the failing tests in `src/lib`

anpa8480:
> > if you revert BotanSymmetricAlgorithm.cpp [8fd89ec](https://github.com/opendnssec/SoftHSMv2/commit/8fd89ec33edfe4a7b510c471101f78b52a0c1294) and uncomment if (mode == SymMode::ECB) all tests are passing.
> > Tested with Botan 2.19.3 image
> 
> That's because ECB is not taken into account during compilation phase. It does not fix the failing tests in `src/lib`

Sorry I dont understand what do you mean "ECB is not taken into account". You mean because Botan_ecb.cpp is missing in CMakelists.txt? The test was performed with dev branch and not fix-botan-ecb

src/lib/test/DeriveTests.cpp
```diff
@@ -703,13 +704,31 @@ void DeriveTests::symDerive(CK_SESSION_HANDLE hSession, CK_OBJECT_HANDLE hKey, C
...
+				CK_BYTE iv[8] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08 };
+				mechEncrypt = { CKM_DES3_CBC, iv, sizeof(iv) };
+			}
+			#endif
 			break;
 		case CKK_AES:
 			mechEncrypt.mechanism = CKM_AES_ECB;
+			#ifdef WITH_BOTAN
+			if (mechType == CKM_AES_CBC_ENCRYPT_DATA) {
+				CK_BYTE iv[16] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16 };
```
zuiderkwast, line 728: I don't know what this is for, but if it's meant to be a continuous sequence from 0x1 to 0x16, then it's wrong. The numbers between 0xa and 0xf are missing.

antoinelochet: I am still investigating why this test does not work on the Encrypt/Decrypt step.

antoinelochet:
I will close this PR since https://github.com/softhsm/SoftHSMv2/pull/771 does a better job than this.

src/lib/test/DeriveTests.cpp
zuiderkwast: Is the just arbitrary data for testing? So it's not as simple as this?

```suggestion
				CK_BYTE iv[16] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10 };
```

antoinelochet: I have tried with only 0x00, and your suggestion. But when decrypting, the bytes 9 to 11 (sometime even the 12th) do not match the corresponding bytes of `data`.

zuiderkwast: Ok, I don't know. I just saw something that looked like a mistake. Ignore me.

Now, the other PR is merged.

antoinelochet:
#771 is merged

src/lib/test/DeriveTests.cpp
antoinelochet: I don't know either. It is a very strange behavior but Encrypt/Decrypt tests work so that must be something I miss.

Issue #729, opened 2023-10-27
openssl operations involving pcks11 and softHSM result in segfault on exit


nhorman:
recently this bug was reported to openssl:
https://github.com/openssl/openssl/issues/22508

Analysis of the problem revealed what is something of an intractable problem.

To summarize the issue:
1) openssl on initalization, loads the pkcs11 library
2) When configured appropriately for this use case, the pkcs11 library loads the SoftHSM library
3) SoftHSM on load creates several global class instances (a SoftHSM object, SecurityManager object, etc)
4) Because SoftHSM is written in C++, the C11 standard mandates that those objects be deleted on exit, so as an implementation detail, the compiler implicitly emits calls to __cxa_atexit, to call the destructor method for each object instance
5) on exit, becuase posix mandates that atexit handlers be run in the reverse order of their registration, the softHSM library deletes the global object instances prior to any previously registered handlers
6) because openssl preforms cleanup in some cases using an atexit handler, it (openssl) makes calls through the pkcs11 engine, which in turn attempts to reference data/code in the softHSM library that has already been deleted via (5), leading to a segfault.

There are a few potential workarounds for this, but it seems to me that the most correct fix would be for softHSM to not delete that data until all references to it were freed from using libraries.  While this is more arguably a shortcoming in the C++ standard, I think the best fix would be for softHSM to modify the implementation of the PCKS api in main.cpp such that it can detect when its global object constructors have been called (via a global variable), and return an appropriate status code without attempting to access the now-deleted object data

nhorman:
FWIW, I believe this issue is related:
https://github.com/opendnssec/SoftHSMv2/issues/722

nhorman:
Its untested as of yet, but something like this I think would avoid the issue:
https://github.com/opendnssec/SoftHSMv2/commit/f45b02cb5b53173a5ee33e0d9294696583bdff61

zandrey:
@nhorman 

FTR: I've tried proposed patch, but it unfortunately brings another issue: SoftHSM throws the error on token creation in free slots.

With the patch applied, and following executed:
```
$ export SOFTHSM2_CONF=/home/zandrey/softhsm2/softhsm2.conf
$ mkdir -p /home/zandrey/softhsm2/softhsm2.tokens
$ softhsm2-util --module /home/zandrey/softhsm2/usr/lib/softhsm/libsofthsm2.so --init-token --free --label test-keys --pin 1111 --so-pin 222222

[...]

PR #742, opened 2024-03-13
Prevent accessing of global c++ objects once they are deleted

Emantor:
Fixes (Maybe) #729.
Reset objects_deleted after reset is called.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -87,6 +87,8 @@
 #include <unistd.h>
 #endif
 
+int objects_deleted = 0;
+
 // Initialise the one-and-only instance
 
 #ifdef HAVE_CXX11
@@ -406,6 +408,8 @@ void SoftHSM::reset()
 {
 	if (instance.get())
 		instance.reset();
+
+	objects_deleted = 0;
 }
 
 // Constructor
@@ -445,6 +449,7 @@ SoftHSM::~SoftHSM()
 
 	isInitialised = false;
 
+	objects_deleted = 1;
 	resetMutexFactoryCallbacks();
 }
 
--- a/src/lib/main.cpp
+++ b/src/lib/main.cpp
@@ -49,6 +49,8 @@
 #define PKCS_API
 #endif
 
+extern int objects_deleted;
+
 // PKCS #11 function list
 static CK_FUNCTION_LIST functionList =
 {
@@ -130,6 +132,8 @@ PKCS_API CK_RV C_Initialize(CK_VOID_PTR pInitArgs)
 {
 	try
 	{
+		if (objects_deleted == 1)
+			return CKR_FUNCTION_FAILED;
 		return SoftHSM::i()->C_Initialize(pInitArgs);
 	}
 	catch (...)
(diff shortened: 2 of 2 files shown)


kraj:
hmm reading through #729 it seems to be a fix for openSSL's behavior where its using `atexit`, I wonder if openSSL exit sequence should not call `atexit` and let the system cleanup happen via system runtime. `atexit` seems to be tied to implementation and for c++ it may run concurrently alongside destructors [1], in C it maybe called concurrently with other functions too.

[1] https://en.cppreference.com/w/cpp/utility/program/atexit

jschlyter:

src/lib/SoftHSM.cpp
```diff
@@ -89,6 +89,8 @@
 
 // Initialise the one-and-only instance
 
+int objects_deleted = 0;
+
 #ifdef HAVE_CXX11
 
 std::unique_ptr<MutexFactory> MutexFactory::instance(nullptr);
```
kalvdans: Don't separate the instance variable from its documentation :)

Emantor: Sorry, I don't get what you mean here.

kalvdans: Sorry to be vague. I think the comment `// Initialise the one-and-only instance` belongs to the definition of `std::unique_ptr<MutexFactory> MutexFactory::instance(nullptr);`.

MorningLightMountain713:
I look forward to this getting merged. Here is a before and after of this patch applied:


Screenshot 2025-01-14 at 1 13 12 PM

src/lib/SoftHSM.cpp
Emantor: I moved the global variable declaration.

bjosv (approved):
Not sure if there is a better way but I guess this work unless we look into some refactoring.
Maybe add a comment with the reason for using a `objects_deleted`.

petrovr:
One flag isInitialised, another flags objects_deleted , tomorrow yet another flag... :( .
And all those changes because third party software "cache sessions".

petrovr:
Almost all issues shows call-stack that starts "at exit", call softhsm ...CloseSession... and then crash i.e., bus error, or segmentation fault, or core dump.

More or less this is expected when a project uses unreliable and non-portable functionality like [at-exit](https://github.com/softhsm/SoftHSMv2/issues/780#issuecomment-2600895708). In addition refuses to proposes portable solution. For instance [cygwin hang](https://github.com/openssl/openssl/issues/10107) is because of "at exit" functionality.

With similar defects third parties try to force project to fix external issues.

May be just to stop processing when is detected "lazy" clean-up request. I mean to check for null pointers in close session.


----

Off topic: Long time ago in pkix-ssh implementation allows use of identity (keys) from engine. This is an openssl loadable module and one thing is clear indirectly - in general module cannot be considered as stateless. It may open database and so  close of database should be triggered as well. It may use network connections, devices , and etc.. Clean shutdown of used resources is preferred.
In consequence  pkix-ssh explicitly clean(free) all used keys and unload modules in main function. Additionally, is performed clean-up of openssl.
With such model "crash" is gone.

nicowilliams:
Do you need to worry about threads?  ISTM that the `objects_deleted` variable needs to be atomic.

PR #761, opened 2024-11-12
Fix 'Assertion failed' in softhsm2-dump-file when encountering a zero-length element

dogo42:
softhsm2-dump-file dies when it encounters a zero-length element
like CKA_SUBJECT in imported keypairs, but tries to treat the
element as std::vector.

Example output:

...
00 00 00 00 00 00 01 01 CKA_SUBJECT
00 00 00 00 00 00 00 03 byte string attribute
00 00 00 00 00 00 00 00 (length 0)
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = unsigned char; _Alloc = std::allocator<unsigned char>; std::vector<_Tp, _Alloc>::reference = unsigned char&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.

Program received signal SIGABRT, Aborted.

### DIFF
--- a/src/bin/dump/softhsm2-dump-file.cpp
+++ b/src/bin/dump/softhsm2-dump-file.cpp
@@ -437,13 +437,15 @@ void dump(FILE* stream)
 			}
 			printf("(length %lu)\n", (unsigned long) len);
 
-			std::vector<uint8_t> value((size_t) len);
-			if (!readBytes(stream, value))
-			{
-				corrupt(stream);
-				return;
+			if (len > 0) {
+				std::vector<uint8_t> value((size_t) len);
+				if (!readBytes(stream, value))
+				{
+					corrupt(stream);
+					return;
+				}
+				dumpBytes(value);
 			}
-			dumpBytes(value);
 		}
 		else if (disktype == ATTRMAP_ATTR)
 		{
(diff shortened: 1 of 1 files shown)


src/bin/dump/softhsm2-dump-file.cpp
```diff
@@ -437,13 +437,15 @@ void dump(FILE* stream)
 			}
 			printf("(length %lu)\n", (unsigned long) len);
 
-			std::vector<uint8_t> value((size_t) len);
-			if (!readBytes(stream, value))
-			{
-				corrupt(stream);
-				return;
+			if (len > 0) {
```
bjosv: Nice. I just have some nits with the placement of the braces
```suggestion
			if (len > 0)
			{
```
which seems to be used in this file.

src/bin/dump/softhsm2-dump-file.cpp
```diff
@@ -461,13 +463,15 @@ void dump(FILE* stream)
...
 			printf("(length %lu)\n", (unsigned long) len);
 
-			std::vector<Attribute> value;
-			if (!readMap(stream, len, value))
-			{
-				corrupt(stream);
-				return;
-			}
-			dumpMap(value);
+			if (len > 0) {
```
bjosv:
```suggestion
			if (len > 0)
			{
```

src/bin/dump/softhsm2-dump-file.cpp
dogo42: Please do stylize as customary in the project, the wrong style came from my brain stem.

src/bin/dump/softhsm2-dump-file.cpp
dogo42: Please do stylize as customary in the project, the wrong style came from my brain stem.

src/bin/dump/softhsm2-dump-file.cpp
bjosv:
```suggestion
			if (len > 0)
			{
```

bjosv: Not sure if the Github GUI lost this change.

bjosv (approved):
LGTM, but with same nitpick comments as `coderabbitai`

PR #762, opened 2024-11-27
fix: Use CMAKE_SHARED_LIBRARY_SUFFIX instead of hardcode .so

middagj:
macOS uses .dylib and this is the only thing that prevents using CMake on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ set(DEFAULT_LOG_LEVEL "INFO"
     CACHE STRING "The default log level")
 set(DEFAULT_OBJECTSTORE_BACKEND "file"
     CACHE STRING "Default storage backend for token objects")
-set(DEFAULT_PKCS11_LIB "${CMAKE_INSTALL_FULL_LIBDIR}/softhsm/libsofthsm2.so"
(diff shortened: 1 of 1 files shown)


jschlyter:
Can you rebase on `develop` so we test results?

middagj:
> Can you rebase on `develop` so we test results?

Done

bjosv:
Nice. We could add CMake builds in CI for both Linux and Mac when this is merged then.

jschlyter (approved):

middagj:
Thanks, I wasn't expecting this quick merge actually. There is another thing that is not helping with macOS (and probably Linux as well). But let me put that in another PR.

middagj:
Never mind, I guess I am just running into the [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/21150) that `install` require relative dirs but `GNUInstallDirs` has absolute paths. Probably the reason why in this project the `CMAKE_INSTALL_SYSCONFDIR` and `CMAKE_INSTALL_LOCALSTATEDIR` are set to absolute values (which is [not recommended](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs)).

The reporter created some alternative you might want to add.

Anyways, this is now how I configure:
```shell
cmake -H. -Bbuild \
    -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR \
    -DCMAKE_INSTALL_SYSCONFDIR=$INSTALL_DIR/etc \
    -DCMAKE_INSTALL_LOCALSTATEDIR=$INSTALL_DIR/var
```

PR #771, opened 2024-11-29
Fix Botan build and test failures

### LINKED ISSUE #767: Fix Botan build failure in CI
Github actions fails for Botan and needs to be fixed before merging pending PRs.

bjosv:
This PR includes the following commits to be able to run all tests on the different platforms in CI.

* Revert removal of Botan 2.0 ECB handling 
    The ECB handling was added when Botan 2.0 was introduced via:
     https://github.com/softhsm/SoftHSMv2/pull/294
    
    but then removed by comments in (and later removed completely):
      https://github.com/softhsm/SoftHSMv2/pull/717 via: https://github.com/softhsm/SoftHSMv2/commit/8fd89ec33edfe4a7b510c471101f78b52a0c1294
    
    This commit reverts parts of this commit to get a stable state.
     We could remove it fully in a separate PR, but then we should also remove the files Botan_ecb.h and Botan_ecb.cpp,
    and avoid running some tests on Botan.

* Correcting build options for Windows in CI 
    The CI installed libraries `Botan` and `OpenSSL` (x86 only) lacks support for `ECC` and `EDDSA`,
    which now are disabled during the CI builds. This fixes the build errors.
    Windows x64 with OpenSSL still builds and runs these tests.

* Remove testing of unsupported `ED448` in Botan 2  
    The test can be re-enabled when we support Botan 3, see https://github.com/randombit/botan/pull/3933
 Today we only support Botan 2 in autoconf/CMake.

Fixes #767

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,12 +67,15 @@ jobs:
           - arch: x64
             backend: openssl
             target-platform: x64
+            build-options:
           - arch: x64
             backend: botan
             target-platform: x64
+            build-options: -DENABLE_ECC=OFF -DENABLE_EDDSA=OFF
           - arch: x86
             backend: openssl
             target-platform: Win32
+            build-options: -DENABLE_ECC=OFF -DENABLE_EDDSA=OFF
     steps:
       - uses: actions/checkout@v4
       - uses: ilammy/msvc-dev-cmd@v1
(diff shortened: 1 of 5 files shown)


jschlyter (approved):

jschlyter:
Nice work!

bjosv:
I'll look into if I can get it faster with caching or something similar, it seems awfully slow..

PR #781, opened 2025-01-21
Add configure option and code to disable OpenSSL engines

bukka:
This is to disable OpenSSL engines that are deprecated and in some setups inconvenient - especially the rdrand default usage which has priority over providers and disallows overwriting random generator using custom providers.

## Summary by CodeRabbit

- **New Features**
	- Added configuration option to disable OpenSSL engines
	- Enhanced control over cryptographic engine initialization

- **Bug Fixes**
	- Improved error handling for cryptographic engine loading
	- Prevented potential null pointer dereferences during engine cleanup

- **Refactor**
	- Modularized OpenSSL engine support with conditional compilation
	- Updated initialization and cleanup processes for cryptographic engines

PR #783, opened 2025-01-28
Update tests to use OpenSSL legacy provider if OpenSSL 3.0 used

bukka:
This enables legacy provider when running tests with OpenSSL 3.0 as some old algorithms are used and it would be good to keep the tested. It also makes tests green when running with OpenSSL 3.0 (except maybe when the crypto RHEL policy is selected).


abbra:
@bukka thanks for this PR. You may want to add [this patch](https://abbra.fedorapeople.org/softhsm-disable-usage-of-openssl-engines.patch) that further constraints engine support to only openssl 1.x. It is from our Fedora patchset, I adjusted it to this PR.

bukka:
@abbra I have actually already created another PR to allow configurable disabling of engines and also skipping them if not available: https://github.com/softhsm/SoftHSMv2/pull/781

jschlyter:
#781 is now merged, please rebase and ensure tests are not failing.

bukka:
Ok I rebased and fix the 1.1.1 build / tests but Ubuntu 24.04 still has got some failure (on both Botan and OpenSSL). The tests work for me with custom OpenSSL builds (tested with latest 3.0 and 3.4) so it might be something Ubuntu 24.04 specific. I will look into it next week.

bukka:
I have been looking into this a bit today after updating my system to Ubuntu 24.04 and I was able to recreate the issue in the pipeline. It seems like it's because of old API usage and some parts that were dropped but need to look into it more into details.

bukka:
Ok this took me quite a bit of time to figure out. Firstly I added change to the assertion to see the actual error codes that did not match. It was actually working fine under gdb and the failed tests were not stable (sometime passed and sometimes fail). So it looked like some sort of race condition which it was at then. After a bit of digging, I tried to disable the ForkTest and it worked. That test basically causing some race when forked process still runs after start of another test which causes those failure. Maybe due to improved scheduling, this is now visible on Ubuntu 24.04 and possibly other newer platforms.

I commented out the test and create a new issue for it to look into it later: https://github.com/softhsm/SoftHSMv2/issues/786

bukka:
@jschlyter This is now ready for review. Everything is rebased and pipeline is fixed including removal of 20.04 Ubuntu runner that is no longer supported. I also applied a proper fix to the ForkTest so this fixes #786 . Commits are independent so if you are happy with it, then best to merge it using a merge commit (as you did for other PR's previously) or using rebase.

sanvila:
Hello @bukka. I am experiencing random test failures in Debian:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1069539

and maybe this PR might fix it.

Regarding the PR itself: The title says "Fix ForkTest" but there is a FIXME in the code saying "This is currently disabled". Is the FIXME really current? Thanks.

bukka:
Hi @sanvila 

> Regarding the PR itself: The title says "Fix ForkTest" but there is a FIXME in the code saying "This is currently disabled". Is the FIXME really current? Thanks.

This was just a left over as previously disabled it. This is now removed as the changes in the ForkTest should fix it - at least they fix it for me and pipeline.

src/lib/test/ObjectTests.cpp
```diff
@@ -94,7 +94,7 @@ void ObjectTests::checkCommonObjectAttributes(CK_SESSION_HANDLE hSession, CK_OBJ
 	};
 
 	rv = CRYPTOKI_F_PTR( C_GetAttributeValue(hSession, hObject, &attribs[0], 1) );
-	CPPUNIT_ASSERT(rv == CKR_OK);
+	CPPUNIT_ASSERT_EQUAL(CKR_OK, rv);
```
kalvdans, line 97: Great that you converted to use a macro that gives better diagnostic outputs! I think the swapped order makes more sense, "assert that the calculated value is equal to the golden reference", but it is not so important. Feel free to create a separate PR with this rewrite to make the actual OpenSSL port shorter.

PR #794, opened 2025-06-16
Implement CKM_RSA_AES_KEY_WRAP mechanism for key wrap/unwrap

ruiliio:
This CKM_RSA_AES_KEY_WRAP implementation enables the wrapping and unwrapping of target asymmetric keys of any length and type using an RSA key. A corresponding test case is included to validate the functionality. 

## Summary by CodeRabbit

- **New Features**
  - Added support for the RSA AES key wrap mechanism, enabling enhanced key wrapping and unwrapping operations.
- **Bug Fixes**
  - Improved correctness of key attribute settings during RSA key generation.
- **Tests**
  - Expanded test coverage to include RSA AES key wrap for various key types and attribute combinations.
  - Enabled additional tests to verify the new key wrapping mechanism.


src/lib/test/AsymWrapUnwrapTests.cpp
```diff
@@ -37,13 +37,14 @@
...
+const CK_BBOOL ON_TOKEN = CK_TRUE;
 const CK_BBOOL IN_SESSION = CK_FALSE;
 
 // CKA_PRIVATE
 const CK_BBOOL IS_PUBLIC = CK_FALSE;
+const CK_BBOOL IS_PRIVATE = CK_TRUE;
 
-
-//CPPUNIT_TEST_SUITE_REGISTRATION(AsymWrapUnwrapTests);
+CPPUNIT_TEST_SUITE_REGISTRATION(AsymWrapUnwrapTests);
```
bjosv, line 47: Oh good finding..the existing test was disabled..

src/lib/SoftHSM.cpp
```diff
@@ -6498,6 +6505,11 @@ CK_RV SoftHSM::C_WrapKey
 			if (rv != CKR_OK)
 				return rv;
 			break;
+		case CKM_RSA_AES_KEY_WRAP:
+			rv = MechParamCheckRSAAESKEYWRAP(pMechanism);
+			if (rv != CKR_OK)
+                return rv;
```
bjosv: Strange indentation, maybe a space to tab change is needed?
(We probably should introduce something like clang-format to avoid having to fix these manually..)

src/lib/SoftHSM.cpp
```diff
@@ -7128,14 +7236,145 @@ CK_RV SoftHSM::C_UnwrapKey
...
 	ByteString wrapped(pWrappedKey, ulWrappedKeyLen);
 	ByteString keydata;
-	if (unwrapKey->getUnsignedLongValue(CKA_CLASS, CKO_VENDOR_DEFINED) == CKO_SECRET_KEY)
-		rv = UnwrapKeySym(pMechanism, wrapped, token, unwrapKey, keydata);
-	else if (unwrapKey->getUnsignedLongValue(CKA_CLASS, CKO_VENDOR_DEFINED) == CKO_PRIVATE_KEY)
-		rv = UnwrapKeyAsym(pMechanism, wrapped, token, unwrapKey, keydata);
+
+	if (pMechanism->mechanism == CKM_RSA_AES_KEY_WRAP)
+	{
+		CK_OBJECT_HANDLE hEmphKey = CK_INVALID_HANDLE;
```
bjosv: Low prio: Could the content of the if-cases be broken out as a separate static function?
Your change follows the current standard I guess, but it would help reading the code.

src/lib/SoftHSM.cpp
```diff
@@ -6712,12 +6726,99 @@ CK_RV SoftHSM::C_WrapKey
 
 	keyClass = wrapKey->getUnsignedLongValue(CKA_CLASS, CKO_VENDOR_DEFINED);
 	ByteString wrapped;
-	if (keyClass == CKO_SECRET_KEY)
-		rv = SoftHSM::WrapKeySym(pMechanism, token, wrapKey, keydata, wrapped);
+
+	if (pMechanism->mechanism == CKM_RSA_AES_KEY_WRAP)
+	{
+		ByteString wrapped_1; // buffer for the wrapped AES key;
```
bjosv: Low prio: Could the content of the if-cases be broken out as a separate static function?
Your change follows the current standard I guess, but it would help reading the code.

ruiliio: Refactored into static functions. Let me know if you’d like any additional changes.

src/lib/SoftHSM.cpp
ruiliio: Refactored into static functions. Let me know if you’d like any additional changes.

src/lib/SoftHSM.cpp
ruiliio: Thanks for your comments. Fixed the indentation.

src/lib/crypto/AsymmetricAlgorithm.h
```diff
@@ -66,6 +66,7 @@ struct AsymMech
 		RSA_MD5_PKCS,
 		RSA_PKCS,
 		RSA_PKCS_OAEP,
+		RSA_AES_KEY_WRAP,
```
kalvdans: Remove this enum value, it doesn't seem to be created anywhere in your patch.

ruiliio: This value is used in AsymmetricAlgorithm::isWrappingMech(AsymMech::Type padding).

kalvdans: Yes, but it is not created anywhere that I can see. By created I mean assigned to a variable or passed as an argument to a function.

src/lib/SoftHSM.cpp
kalvdans: Technically, they are member functions. Looks better to me at least, thanks for breaking them out.

src/lib/crypto/AsymmetricAlgorithm.h
ruiliio: You are right! Thank you for pointing it out. They are removed in the latest commit.

jschlyter (approved):

PR #799, opened 2025-07-22
Add optional enable_64bit_file_offset

LawrenceHunter:
Fixes an error when running a 32 bit version of SoftHSM where readdir fails with errno 75.


LawrenceHunter:
> Just unconditionally enable it, there's no reason to use 32-bit file offsets even on 32-bit systems.

Okay I'll set it enabled by default

kalvdans:
> > Just unconditionally enable it, there's no reason to use 32-bit file offsets even on 32-bit systems.
> 
> Okay I'll set it enabled by default

No, don't even make it configurable. Hardcode it in CFLAGS.

bjosv (commented):
Nice.
Either remove the AC_ARG_ENABLE part, or just add the flags in configure.ac directly, like:
```
# Compiler flags
CFLAGS="${CFLAGS} -D_FILE_OFFSET_BITS=64"
CXXFLAGS="${CXXFLAGS} -D_FILE_OFFSET_BITS=64"
ACX_PEDANTIC
ACX_STRICT
ACX_64BIT
AX_CXX_COMPILE_STDCXX_11([noext],[optional])
```

A similar change for CMake should fit in `cmake/modules/CompilerOptions.cmake`.

LawrenceHunter:
Thanks for the advice both, it is now included in the flags.

kalvdans (approved):

bjosv (approved):

PR #802, opened 2025-07-27
Replace deprecated Github runner ubuntu-20.04 

bjosv:
Replace deprecated Github runner `ubuntu-20.04` with `ubuntu-24.04`

When running with a newer ubuntu a new issue surfaces with the fork test,
which probably is due to the lift of GLIBC, from v2.31 to v2.34.
Others have [this](https://sourceware.org/bugzilla/show_bug.cgi?id=4737) related issue.

An fix for this is to wait for spawned fork-tests to avoid any race conditions.
Each spawned testcase is now exited to avoid running the tests multiple times.
    

## Summary by CodeRabbit

* **Chores**
  * Updated CI workflow to use Ubuntu 24.04 and ensured compatibility by installing required OpenSSL packages.

* **Tests**
  * Improved process handling and validation in fork-related tests to ensure proper cleanup and verification of child process completion.

### DIFF
--- a/src/lib/test/ForkTests.cpp
+++ b/src/lib/test/ForkTests.cpp
@@ -40,6 +40,7 @@
 #include "osmutex.h"
 
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 CPPUNIT_TEST_SUITE_REGISTRATION(ForkTests);
@@ -64,6 +65,7 @@ void ForkTests::testFork()
 {
 	CK_RV rv;
 	pid_t pid;
+	int status;
 
 	// Just make sure that we finalize any previous failed tests
 	CRYPTOKI_F_PTR( C_Finalize(NULL_PTR) );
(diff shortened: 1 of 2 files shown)


bukka:
Although it might be still good idea to have 1.1.1 in the pipeline. Maybe merging our approaches together would be best. In that case it could be added to Ubuntu 22.04 runner and 24.04 would stay to test OpenSSL 3.0.

src/lib/test/ForkTests.cpp
```diff
@@ -78,12 +80,20 @@ void ForkTests::testFork()
...
+			rv = CRYPTOKI_F_PTR( C_Finalize(NULL_PTR) );
+			CPPUNIT_ASSERT(rv == CKR_OK);
+			_exit(EXIT_SUCCESS);
 			break;
 		default:
+			/* For the parent, the token is expected to still be initialized. */
 			rv = CRYPTOKI_F_PTR( C_Initialize(NULL_PTR) );
 			CPPUNIT_ASSERT(rv == CKR_CRYPTOKI_ALREADY_INITIALIZED);
+			/* Wait for the child process to finish and check its status. */
+			CPPUNIT_ASSERT(waitpid(pid, &status, 0) == pid);
```
bukka, line 95: it actually doesn't need to wait for the child...

kalvdans: The unit test must wait for the child, otherwise errors raised in the child will not be propagated back to the testing framework.

I think these changes to testing looks valid. We shouldn't have a fork bomb :)

They belong in a separate PR though, this one is about replacing github runner.

bukka: Ah yeah that's actually good point. It tests the child exit status below so this is actually better.

jschlyter (approved):

PR #806, opened 2025-08-11
Update tests to use OpenSSL legacy provider if OpenSSL 3.0 used

bukka:
Also added pipeline to test OpenSSL 3.0 and removed matrix which didn't really make much sense and was just adding unnecessary deps for each workflow.


src/lib/test/p11test.cpp
```diff
@@ -71,6 +78,11 @@ int main(int /*argc*/, char**const /*argv*/)
 #endif
 #endif
 
+#if defined(WITH_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x3000000fL
+	OSSL_PROVIDER_load(NULL, "legacy");
+	OSSL_PROVIDER_load(NULL, "default");
+#endif
```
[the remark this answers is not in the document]
bukka: Such check is pointless in this case as those providers won't fail and it's just a test anyway...

src/lib/crypto/test/cryptotest.cpp
```diff
@@ -75,6 +78,11 @@ std::auto_ptr<BotanCryptoFactory> BotanCryptoFactory::instance(NULL);
 
 int main(int /*argc*/, char** /*argv*/)
 {
+#if defined(WITH_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x3000000fL
+	OSSL_PROVIDER_load(NULL, "legacy");
+	OSSL_PROVIDER_load(NULL, "default");
+#endif
```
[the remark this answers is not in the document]
bukka: Such check is pointless in this case as those providers won't fail and it's just a test anyway...

src/lib/crypto/test/cryptotest.cpp
```diff
@@ -75,6 +78,11 @@ std::auto_ptr<BotanCryptoFactory> BotanCryptoFactory::instance(NULL);
 
 int main(int /*argc*/, char** /*argv*/)
 {
+#if defined(WITH_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x3000000fL
```
kalvdans: Off by a zero (man openssl_version_number says the format is MMNNFFPPS: major minor fix patch status) and don't use floating point comparison.

```suggestion
#if defined(WITH_OPENSSL) && OPENSSL_VERSION_NUMBER >= 0x30000000
```

bukka: Ah yeah, fixed, thanks

bukka: It's not floating point comparison btw :) . It's hex last version. Not really an issue though but it makes more sense to use 0...

bukka: So basically I just didn't include pre-releases in 3.0 as it's defined as:

```
/* Synthesize OPENSSL_VERSION_NUMBER with the layout 0xMNN00PPSL */
# ifdef OPENSSL_VERSION_PRE_RELEASE
#  define _OPENSSL_VERSION_PRE_RELEASE 0x0L
# else
#  define _OPENSSL_VERSION_PRE_RELEASE 0xfL
# endif
# define OPENSSL_VERSION_NUMBER          \
    ( (OPENSSL_VERSION_MAJOR<<28)        \
      |(OPENSSL_VERSION_MINOR<<20)       \
      |(OPENSSL_VERSION_PATCH<<4)        \
      |_OPENSSL_VERSION_PRE_RELEASE )
```

bukka: But I can see that it might be confusing for people so changed to the standard way... :)

kalvdans: Aha! My bad, then number of digits are okay. Sorry for the noise!

.github/workflows/ci.yml
```diff
@@ -32,10 +46,31 @@ jobs:
...
+        run: |
+          ./autogen.sh
+          ./configure --with-crypto-backend=openssl
+          make
+      - name: Test
+        run: |
+          make check || (find . -name test-suite.log -exec cat {} \; && false)
+
+  linux_ossl_30:
+    name: Linux (openssl 3.0)
```
bjosv: Might look a bit nicer on the Github job summary (i.e. similar to 1.1.1 above):
```suggestion
    name: Linux with OpenSSL 3.0
```

src/lib/test/Makefile.am
```diff
@@ -1,9 +1,11 @@
 MAINTAINERCLEANFILES = 		$(srcdir)/Makefile.in
 
-AM_CPPFLAGS = 			-I$(srcdir)/.. \
+AM_CPPFLAGS = 	-I$(srcdir)/../.. \
```
bjosv: If this extra path needed? Not sure if there is anything needed from `<repo root>/src/`
If its for config.h from `<repo root>` then that should be taken care by autoconf via AC_CONFIG_HEADERS and the generated DEFAULT_INCLUDES.
I cant see that the other tests have this added either.

bukka: Yeah it's for `#include "config.h"` in `src/lib/test/p11test.cpp` to be able to use `WITH_OPENSSL`. It doesn't compile without this change.

This is done in the same way like it is in https://github.com/softhsm/SoftHSMv2/blob/76c7ed80f158fea3e0a3d13534c692aca01817aa/src/lib/crypto/test/Makefile.am#L3-L4

.github/workflows/ci.yml
bukka: Ah yeah forgot to change this. Fixed. Thanks.

src/lib/test/Makefile.am
bjosv: Strange that is makes a difference since `I$(srcdir)/../..` should become `I./../.. `,
which would be including `src/` when looking from `src/lib/test/`. 
To me it should rather be `-I$(srcdir)/../../..` to be able to find `config.h` in the repo root.
Maybe its a difference in our automake versions or something..

bukka: Are you saying that it works for you without this change?

bjosv: Yes, in my build I'll get the following compile command without the added include, which works here.
```
make[4]: Entering directory '/home/bjorn/repos/SoftHSMv2/src/lib/test'
g++ -std=c++11 -DHAVE_CONFIG_H -I. -I../../..  -I./.. -I./../common -I./../pkcs11    -g -O2 -D_FILE_OFFSET_BITS=64 -Wall -Wextra -fvisibility=hidden -MT p11test.o -MD -MP -MF $depbase.Tpo -c -o p11test.o p11test.cpp &&\
```
The generated Makefile got `DEFAULT_INCLUDES = -I. -I$(top_builddir)` that is added as the first include directives, which would be needed to find config.h
Maybe it's a classic "it works on my machine" :) so If you need it in your setup then I'm happy keeping it.

bukka: Hmm so I just retried and it normally works. I actually accidentlally removed `I$(srcdir)/..` as well but even that one does not seem to be needed as it works normally without it.

bukka: I had this issue originally on Ubuntu 20.04 but since then I updated to 24.04 so it might have been older version of the tools. Or I had some messed up build.

.github/workflows/ci.yml
```diff
@@ -32,10 +46,31 @@
...
+
+  linux_ossl_30:
+    name: Linux with OpenSSL 3.0
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Prepare
+        run: |
+          sudo apt update -qq
+          sudo apt install libcppunit-dev p11-kit
```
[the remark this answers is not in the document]
bukka: Seems to be already installed in the image as it works...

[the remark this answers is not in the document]
bukka: Anyway I added it.

src/lib/test/Makefile.am
bukka: Pushed to fixed version.

bjosv: Ah, great that we see the same.

PR #809, opened 2025-08-18
Added ML-DSA support

antoinelochet:
Adds ML-DSA support with OpenSSL >= 3.5

## Summary by CodeRabbit

* **New Features**
  * ML‑DSA support: key‑pair generation, signing/verification, parameter sets, key objects, wrapping/unwrapping, OpenSSL backend, and mechanism parameters (hedge/context); session mechanism‑param handling added.

* **Build & Configuration**
  * New option to enable ML‑DSA with corresponding compile‑time macros and config entries.

* **CI**
  * Added Ubuntu 24.04 workflow testing with OpenSSL 3.5.5.

* **Tests**
  * New ML‑DSA test suite covering generation, serialization/PKCS#8, signing/verification and edge cases.

* **Documentation**
  * README and CMake notes updated with the ML‑DSA configuration option.


src/lib/SoftHSM.cpp
```diff
@@ -12602,6 +13001,66 @@ CK_RV SoftHSM::getEDPublicKey(EDPublicKey* publicKey, Token* token, OSObject* ke
 	return CKR_OK;
 }
 
+CK_RV SoftHSM::getMLDSAPrivateKey(MLDSAPrivateKey* privateKey, Token* token, OSObject* key)
+{
```
kalvdans: Is the implicit `this` pointer used here? If not, you can mark this function static or move it out of the class.

same for getMLDSAPublicKey and setMLDSAPrivateKey.

antoinelochet: I took example on other methods like `setRSAPrivateKey` or `getRSAPrivateKey` which are not static.
I do not know if it is better to change the style of these methods.
But I agree that moving them out will unclutter `SoftHSM.cpp`

kalvdans: No need to follow bad examples, in my opinion. We can fix the other functions in the future.

antoinelochet: Which approach is preferable ?

m4/acx_openssl_mldsa.m4
```diff
@@ -0,0 +1,43 @@
...
+		],[
+			AC_MSG_RESULT([yes])
+			acx_cv_lib_openssl_mldsa_support=yes
+		],[
+			AC_MSG_RESULT([no])
+			acx_cv_lib_openssl_mldsa_support=no
+		],[
+			AC_MSG_WARN([Cannot test, ML-DSA])
+			acx_cv_lib_openssl_mldsa_support=no
+		])
```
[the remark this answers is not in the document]
antoinelochet: No need to check version when context is enough

src/lib/crypto/MLDSAParameters.h
```diff
@@ -0,0 +1,74 @@
...
+
+	/* ML-DSA values for CKA_PARAMETER_SETS
+	typedef CK_ULONG CK_ML_DSA_PARAMETER_SET_TYPE;
+	#define CKP_ML_DSA_44          0x00000001UL
+	#define CKP_ML_DSA_65          0x00000002UL
+	#define CKP_ML_DSA_87          0x00000003UL
+	 */
+	static const unsigned long ML_DSA_44_PARAMETER_SET = CKP_ML_DSA_44;
+	static const unsigned long ML_DSA_65_PARAMETER_SET = CKP_ML_DSA_65;
+	static const unsigned long ML_DSA_87_PARAMETER_SET = CKP_ML_DSA_87;
```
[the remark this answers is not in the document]
antoinelochet: I find it clearer to leave the PKCS#11 constants

src/lib/SoftHSM.cpp
kalvdans: I like top-level functions rather than static methods, but it's just taste. Marking it static will be less intrusive in the current patch.

antoinelochet: I have moved the methods to a new util class: `MLDSAUtil.cpp` on the model of `OSSLUtil.cpp`.
I am not sure of the calss location, maybe it should be alongside `SoftHSM.cpp`.
Please let me know what you think of it.

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -0,0 +1,467 @@
...
+		return false;
+	}
+	if (!EVP_DigestSign(ctx, &signature[0], &len, dataToSign.const_byte_str(), dataToSign.size()))
+	{
+		ERROR_MSG("ML-DSA sign failed (0x%08X)", ERR_get_error());
+		EVP_MD_CTX_free(ctx);
+		return false;
+	}
+	EVP_MD_CTX_free(ctx);
+	return true;
```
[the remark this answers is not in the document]
antoinelochet: Already done line 73

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -0,0 +1,467 @@
...
+	int initRv = EVP_PKEY_verify_message_init(vctx, sig_alg, NULL);
+	if (!initRv) {
+		ERROR_MSG("ML-DSA verify init failed (0x%08X)", ERR_get_error());
+		EVP_PKEY_CTX_free(vctx);
+		EVP_SIGNATURE_free(sig_alg);
+		return false;
+	}
+	int verifyRV = EVP_PKEY_verify(vctx, signature.const_byte_str(), signature.size(),
+                                            originalData.const_byte_str(), originalData.size());
+
```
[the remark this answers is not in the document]
antoinelochet: EVP_PKEY_verify_message does not exist

m4/acx_botan_mldsa.m4
```diff
@@ -0,0 +1,33 @@
...
+			#endif
+			int main(void){ return 0; }
+			])
+		], [
+			acx_cv_lib_botan_mldsa_support=yes
+			AC_MSG_RESULT([yes])
+		], [
+			acx_cv_lib_botan_mldsa_support=no
+			AC_MSG_RESULT([no])
+		])
```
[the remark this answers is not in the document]
antoinelochet: Won't change

src/lib/SoftHSM.cpp
```diff
@@ -9873,6 +10013,265 @@ CK_RV SoftHSM::generateED
...
+	{
+		switch (pPublicKeyTemplate[i].type)
+		{
+			case CKA_PARAMETER_SET:
+				if (pPublicKeyTemplate[i].ulValueLen != sizeof(CK_ULONG)) {
+                    INFO_MSG("CKA_PARAMETER_SET must be sizeof(CK_ULONG)");
+                    return CKR_ATTRIBUTE_VALUE_INVALID;
+                }
+                paramSet = *(CK_ULONG*)pPublicKeyTemplate[i].pValue;
+				break;
```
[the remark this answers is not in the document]
antoinelochet: Changes done

src/lib/SoftHSM.cpp
```diff
@@ -6292,23 +6416,23 @@ CK_RV SoftHSM::WrapKeySym
...
 			break;
-			
+
 		case CKM_DES3_CBC_PAD:
 			blocksize = 8;
 			wrappedlen = RFC5652Pad(keydata, blocksize);
 			algo = SymAlgo::DES3;
 			break;
-			
+
```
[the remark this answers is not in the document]
antoinelochet: Ignore

src/lib/SoftHSM.cpp
```diff
@@ -140,6 +146,8 @@ static CK_RV newP11Object(CK_OBJECT_CLASS objClass, CK_KEY_TYPE keyType, CK_CERT
 				*p11object = new P11GOSTPublicKeyObj();
 			else if (keyType == CKK_EC_EDWARDS)
 				*p11object = new P11EDPublicKeyObj();
+			else if (keyType == CKK_ML_DSA)
+				*p11object = new P11MLDSAPublicKeyObj();
 			else
```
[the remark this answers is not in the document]
antoinelochet: Applied

src/lib/SoftHSM.cpp
```diff
@@ -4143,6 +4189,9 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
 #endif
 #ifdef WITH_EDDSA
 	bool isEDDSA = false;
+#endif
+#ifdef WITH_ML_DSA
+	bool isMLDSA = false;
 #endif
 	switch(pMechanism->mechanism) {
```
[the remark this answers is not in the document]
antoinelochet: Parameter set is inferred from key length

src/lib/crypto/OSSLMLDSAPrivateKey.cpp
```diff
@@ -0,0 +1,194 @@
...
+	ByteString valueBS = ByteString(priv, priv_len);
+
+	setValue(valueBS);
+}
+
+// Check if the key is of the given type
+bool OSSLMLDSAPrivateKey::isOfType(const char* inType)
+{
+	return !strcmp(type, inType);
+}
```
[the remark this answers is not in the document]
antoinelochet: ignore

src/lib/crypto/OSSLUtil.cpp
```diff
@@ -34,10 +34,20 @@
...
 #include "DerUtil.h"
 #include "OSSLUtil.h"
+#include "MLDSAParameters.h"
+#include <map>
 #include <openssl/asn1.h>
 #include <openssl/evp.h>
 #include <openssl/err.h>
 
+#ifdef WITH_ML_DSA
+static std::map<unsigned long, const char*> mldsaAlgNameFromParameterSet {
```
kalvdans:
```suggestion
static const std::map<unsigned long, const char*> mldsaAlgNameFromParameterSet {
```

src/lib/SoftHSM.cpp
kalvdans: Looks good, no more unused implicit `this` parameter!

We could have a look at C++ `namespace` feature to group functions together without using a class.

antoinelochet: I have tried using `namespace` but could not make it work unfortunately.
My C++ is not as good as my Java :)
Thanks for your input !

src/lib/SoftHSM.cpp
```diff
@@ -6292,23 +6421,23 @@ CK_RV SoftHSM::WrapKeySym
...
 			break;
-			
+
 		case CKM_DES3_CBC_PAD:
 			blocksize = 8;
 			wrappedlen = RFC5652Pad(keydata, blocksize);
 			algo = SymAlgo::DES3;
 			break;
-			
+
```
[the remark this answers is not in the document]
antoinelochet: ignore

src/lib/pkcs11/pkcs11.h
```diff
@@ -893,6 +897,38 @@ typedef unsigned long ck_mechanism_type_t;
 #define CKM_EC_EDWARDS_KEY_PAIR_GEN	(0x1055UL)
 #define CKM_EDDSA			(0x1057UL)
 
+/* From version 3.2 */
```
bukka: But this header still advertise 2.40.0. I think it should update the whole header.

m4/acx_botan_mldsa.m4
```diff
@@ -0,0 +1,33 @@
+AC_DEFUN([ACX_BOTAN_MLDSA],[
```
bukka: Shouldn't this be added once Botan support is added?

src/lib/pkcs11/pkcs11.h
bukka: The issue is that p11-kit has only update for 3.1.0 so maybe bum that first and then add those temporary changes with bump to 3.2.0

bukka: https://github.com/p11-glue/p11-kit/blob/master/common/pkcs11.h

bukka: PR for the header updated created in https://github.com/softhsm/SoftHSMv2/pull/815

m4/acx_botan_mldsa.m4
antoinelochet: Removed and error message on configure when `--enable-mldsa` is passed on the command line

primetomas:
Very cool. Did some manual test.

I checked out this branch and tested with our p11ng-cli from EJBCA (Java code using Bouncy Castle). Generating keys, listing keys, signing (and verifying the signatures with Bouncy Castle code). I have PKCS#11v3.2 draft implemented and it worked out-of-the-box. Works well with all variants ML-DSA-44, ML-DSA-65 and ML-DSA-87.

jschlyter:
Nice work! I'll ask @bjosv to review before merge.

src/lib/crypto/OSSLMLDSAPublicKey.cpp
```diff
@@ -0,0 +1,147 @@
...
+		ERROR_MSG("Could not EVP_PKEY_fromdata_init:%d", rv);
+		EVP_PKEY_CTX_free(ctx);
+		return;
+	}
+	rv = EVP_PKEY_fromdata(ctx, &pkey, selection, params);
+	if (!rv) {
+		ERROR_MSG("Could not EVP_PKEY_fromdata:%d", rv);
+		EVP_PKEY_CTX_free(ctx);
+		return;
+	}
```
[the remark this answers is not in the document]
antoinelochet: Ignore

src/lib/pkcs11/pkcs11.h
antoinelochet: I think I will let the 2.4 version for this PR and will let the mainteners decide when they want to update the supported PKCS#11 version.
I have a development for ML-KEM and I think this would be more criticial since new functions are added.

primetomas:
I found an issue. It does not handle signing with a provided ML-DSA context. I.e. PKCS11v3.2 section 6.67.5.

`It has an optional parameter CK_SIGN_ADDITIONAL_CONTEXT. If no parameter is supplied the hedgeVariant will be CKH_HEDGE_PREFERRED, ulContextLen will be zero and pContext will be NULL. On signing, if hedgeVariant is set to CKH_HEDGE_PREFERRED, the token may create either a hedged signature or a deterministic signature as specified in [FIPS 204]. If hedgeVariant is set to CKH_HEDGE_REQUIRED, the token must produce a hedged signature or fail. If the hedgeVariant is set to CKH_DETERMINISTIC_REQUIRED, the token must produce a deterministic signature or fail. On verification the hedgeVariant parameter is ignored.`

Referring to section 6.67.1:
```
CK_SIGN_ADDITIONAL_CONTEXT is used in the mechanism parameters to supply a NIST defined context string in signature scheme.
typedef struct CK_SIGN_ADDITIONAL_CONTEXT {
  CK_HEDGE_TYPE  hedgeVariant;
  CK_BYTE_PTR    pContext;
  CK_ULONG       ulContextLen;
} CK_SIGN_ADDITIONAL_CONTEXT;
;
```

This is needed for Composite certificates as a domain separator is used as specified in [composite draft 12 section 7](https://datatracker.ietf.org/doc/html/draft-ietf-lamps-pq-composite-sigs-12#sec-alg-parms).

antoinelochet:
> I found an issue. It does not handle signing with a provided ML-DSA context. I.e. PKCS11v3.2 section 6.67.5.
> 
> `It has an optional parameter CK_SIGN_ADDITIONAL_CONTEXT. If no parameter is supplied the hedgeVariant will be CKH_HEDGE_PREFERRED, ulContextLen will be zero and pContext will be NULL. On signing, if hedgeVariant is set to CKH_HEDGE_PREFERRED, the token may create either a hedged signature or a deterministic signature as specified in [FIPS 204]. If hedgeVariant is set to CKH_HEDGE_REQUIRED, the token must produce a hedged signature or fail. If the hedgeVariant is set to CKH_DETERMINISTIC_REQUIRED, the token must produce a deterministic signature or fail. On verification the hedgeVariant parameter is ignored.`
> 
> Referring to section 6.67.1:
> 
> ```
> CK_SIGN_ADDITIONAL_CONTEXT is used in the mechanism parameters to supply a NIST defined context string in signature scheme.
> typedef struct CK_SIGN_ADDITIONAL_CONTEXT {
>   CK_HEDGE_TYPE  hedgeVariant;
>   CK_BYTE_PTR    pContext;
>   CK_ULONG       ulContextLen;
> } CK_SIGN_ADDITIONAL_CONTEXT;
> ;
> ```
> 
> This is needed for Composite certificates as a domain separator is used as specified in [composite draft 12 section 7](https://datatracker.ietf.org/doc/html/draft-ietf-lamps-pq-composite-sigs-12#sec-alg-parms).

I think there is (was?) an issue with OpenSSL. I will have to check.
Do you happen to have a test case ?

src/lib/crypto/test/MLDSATests.cpp
```diff
@@ -0,0 +1,506 @@
+/*****************************************************************************
+ EDDSATests.cpp
+
+ Contains test cases to test the EDDSA class
+ *****************************************************************************/
+
+#include <stdlib.h>
```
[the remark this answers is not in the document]
antoinelochet: ignore

antoinelochet:
> > I think there is (was?) an issue with OpenSSL. I will have to check. Do you happen to have a test case ?
> 
> I have a test, but in Java using Bouncy Castle (for the Composite certs) and an internal PKCS11 Provider.

I have added the support for `CK_SIGN_ADDITIONAL_CONTEXT`.
Can you check that it is ok with your tests ?

src/lib/SoftHSM.cpp
```diff
@@ -5420,6 +5545,54 @@ CK_RV SoftHSM::AsymVerifyInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMech
...
+						case CKH_HEDGE_PREFERRED:
+						default:
+							additionalContext.hedgeType = Hedge::HEDGE_PREFERRED;
+					}
+					param = &additionalContext;
+					paramLen = sizeof(additionalContext);
+				}
+			}
+			break;
 #endif
```
antoinelochet, line 5591: In src/lib/SoftHSM.cpp around lines 5550 to 5596, the ML-DSA handling allocates
additionalContext.contextAsChar with malloc() but only copies the
SIGN_ADDITIONAL_CONTEXT struct elsewhere, leaving the malloc'd buffer orphaned
on return; free that malloc'd contextAsChar before every return path in both
AsymSignInit and AsymVerifyInit (or change the ownership so the session makes a
deep copy of the context bytes) to ensure the allocated buffer is either
transferred correctly or freed on all error/success exits.

src/lib/SoftHSM.cpp
```diff
@@ -6352,23 +6565,23 @@ CK_RV SoftHSM::WrapKeySym
...
 			algo = SymAlgo::AES;
 			break;
-			
+
 		case CKM_DES3_CBC:
 			algo = SymAlgo::DES3;
 			break;
-			
+
 		case CKM_DES3_CBC_PAD:
```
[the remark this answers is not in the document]
antoinelochet: ignore

src/lib/crypto/OSSLMLDSAKeyPair.h
```diff
@@ -0,0 +1,42 @@
...
+ OSSLMLDSAKeyPair.h
+
+ OpenSSL ML-DSA key-pair class
+ *****************************************************************************/
+
+#ifdef WITH_ML_DSA
+#ifndef _SOFTHSM_V2_OSSLMLDSAKEYPAIR_H
+#define _SOFTHSM_V2_OSSLMLDSAKEYPAIR_H
+
+#include "config.h"
```
kalvdans: Please include "config.h" before testing WITH_ML_DSA, as it is config.h that's supposed to define it.

src/lib/crypto/OSSLUtil.cpp
```diff
@@ -34,10 +34,22 @@
...
 #include <openssl/evp.h>
 #include <openssl/err.h>
 
+#ifdef WITH_ML_DSA
+static const std::map<unsigned long, const char*> mldsaAlgNameFromParameterSet {
+	{MLDSAParameters::ML_DSA_44_PARAMETER_SET, "ML-DSA-44"},
+	{MLDSAParameters::ML_DSA_65_PARAMETER_SET, "ML-DSA-65"},
+	{MLDSAParameters::ML_DSA_87_PARAMETER_SET, "ML-DSA-87"}
+};
+#endif
```
kalvdans, line 51: Move this lookup table to inside OSSL::mldsaParameterSet2Name where the only place it is used.

src/lib/session_mgr/Session.cpp
```diff
@@ -352,6 +352,41 @@ void Session::setParameters(void* inParam, size_t inParamLen)
 		paramLen = 0;
 	}
 
+	// For ML-DSA, deep-copy struct + context bytes into one buffer and fix the pointer
+	// Suggested by CodeRabbit to avoid memory leak
+    if (mechanism == AsymMech::MLDSA && inParamLen == sizeof(SIGN_ADDITIONAL_CONTEXT)) {
```
kalvdans: It sounds fragile to determine which struct we are getting only depending on the size of the struct. Please design a more robust interface where the caller explicitly say that a SIGN_ADDITIONAL_CONTEXT struct is passed to this function.

src/lib/P11Objects.h
```diff
@@ -198,6 +198,19 @@ class P11EDPublicKeyObj : public P11PublicKeyObj
...
+{
+public:
+	// Constructor
+	P11MLDSAPublicKeyObj();
+
+	// Add attributes
+	virtual bool init(OSObject *inobject);
+
+protected:
+	bool initialized;
```
kalvdans, line 211: Describe why we need to keep initialization status in an instance variable. Isn't the object initialized after the constructor returns?

src/lib/test/SignVerifyTests.cpp
```diff
@@ -803,6 +846,97 @@ void SignVerifyTests::testEdSignVerify(const char* curve)
...
+		{ CKH_HEDGE_PREFERRED,  NULL,   0  },
+		{ CKH_HEDGE_PREFERRED,  data,   dataSize  },
+		{ CKH_HEDGE_REQUIRED,  NULL,   0  },
+		{ CKH_HEDGE_REQUIRED,  data,   dataSize  },
+		{ CKH_DETERMINISTIC_REQUIRED,  NULL,   0  },
+		{ CKH_DETERMINISTIC_REQUIRED,  data,   dataSize  },
+	};
+
+	// Public Session keys
+	rv = generateMLDSA(parameterSet,hSessionRW,IN_SESSION,IS_PUBLIC,IN_SESSION,IS_PUBLIC,hPuk,hPrk);
```
kalvdans, line 895: Please put a space after comma, to make the code easier to read. (more occurrances in this file)

src/lib/P11Objects.h
antoinelochet: This is how all the other objects are coded in this file. Do you have a change suggestion ?

src/lib/crypto/OSSLMLDSAKeyPair.h
antoinelochet: Changed

src/lib/test/SignVerifyTests.cpp
antoinelochet: Changed

src/lib/P11Objects.h
kalvdans: Initialize the object in the constructor, to avoid the possibility of a half-initialized object lying around.

src/lib/crypto/OSSLMLDSAKeyPair.h
kalvdans: Same for src/lib/crypto/test/MLDSATests.cpp (please double-check all files)

src/lib/test/SignVerifyTests.cpp
kalvdans: oh, thanks, but now the whole file got reformatted :(

I thought only to write your newly added code in a readable fashion.

It's better to do pure reformatting as a separate PR, to make this PR cleaner and easier to review.

kalvdans: Actually, github has a feature to ignore whitespace when viewing a diff, so it's not at all a big problem. :)

antoinelochet: Should be better now.

src/lib/P11Objects.h
antoinelochet: I think this could be done in an later PR.
This has the same structure as all of the other classes. I like the coherence.

src/lib/session_mgr/Session.cpp
antoinelochet: I do not know if adding a specific ML-DSA method on this class would be acceptable.
Moreover, SIGN_ADDITIONAL_CONTEXT is the only param that is expected for ML-DSA.
I think that there is not much I can do here expect a new method.
What is your opinion on this @softhsm/core ?

src/lib/SoftHSM.cpp
```diff
@@ -9935,6 +10170,266 @@ CK_RV SoftHSM::generateED
...
+			OSObject* ospub = (OSObject*)handleManager->getObject(*phPublicKey);
+			handleManager->destroyObject(*phPublicKey);
+			if (ospub) ospub->destroyObject();
+			*phPublicKey = CK_INVALID_HANDLE;
+		}
+	}
+
+	return rv;
+}
+#endif
```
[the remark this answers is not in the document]
antoinelochet: Ignore

src/lib/SoftHSM.cpp
```diff
@@ -9935,6 +10170,266 @@ CK_RV SoftHSM::generateED
...
+	{
+		switch (pPublicKeyTemplate[i].type)
+		{
+			case CKA_PARAMETER_SET:
+				if (pPublicKeyTemplate[i].ulValueLen != sizeof(CK_ULONG)) {
+                    INFO_MSG("CKA_PARAMETER_SET must be sizeof(CK_ULONG)");
+                    return CKR_ATTRIBUTE_VALUE_INVALID;
+                }
+                paramSet = *(CK_ULONG*)pPublicKeyTemplate[i].pValue;
+				break;
```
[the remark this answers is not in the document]
antoinelochet: Ignore

src/lib/crypto/test/MLDSATests.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+	SIGN_ADDITIONAL_CONTEXT context = {
+		Hedge::Type::DETERMINISTIC_REQUIRED,
+		NULL,
+		0
+	};
+
+	// Sign the data
+	ByteString sig;
+	CPPUNIT_ASSERT(mldsa->sign(kp->getPrivateKey(), dataToSign, sig, AsymMech::MLDSA, &context, sizeof(context)));
```
bukka: It would be good to verify that sign produces something as passing this directly to verify does not actually test that things are completely working (e.g. if both sign and verify are wrong). As this is deterministic, you could at least do something like (pseudocode - it would need of course compare bytes - not pointers but you get the idea):

```
ByteString sig1, sig2;
CPPUNIT_ASSERT(mldsa->sign(..., sig1, ..., &context, sizeof(context)));
CPPUNIT_ASSERT(mldsa->sign(..., sig2, ..., &context, sizeof(context)));
CPPUNIT_ASSERT(sig1 == sig2);
```

Also we should probably add some known static signatures for verify. So we test it agains real signature and not only from result of sign...

src/lib/crypto/MLDSAPrivateKey.cpp
```diff
@@ -0,0 +1,98 @@
...
+const ByteString& MLDSAPrivateKey::getSeed() const
+{
+	return seed;
+}
+
+// Serialisation
+ByteString MLDSAPrivateKey::serialise() const
+{
+	return seed.serialise() +
+	       value.serialise();
```
bukka, line 83: NIT: you could keep on one line... :)

src/lib/session_mgr/Session.cpp
```diff
@@ -352,6 +352,41 @@ void Session::setParameters(void* inParam, size_t inParamLen)
...
+            unsigned char* payload = buf + sizeof(SIGN_ADDITIONAL_CONTEXT);
+            memcpy(payload, additionalContext->contextAsChar, ctxLen);
+            stored->contextAsChar = payload;
+        } else {
+            stored->contextAsChar = NULL;
+        }
+        param  = buf;
+        paramLen = sizeof(SIGN_ADDITIONAL_CONTEXT);
+        return;
+    }
```
bukka: I think this logic should go to `AsymSignInit` and `AsymVerifyInit`. It should probably also go to some common static function there as the logic seems pretty much the same (haven't checked if there are any differences but if they are, then they are small to be handled in that common function).

src/lib/SoftHSM.cpp
```diff
@@ -1026,6 +1047,19 @@ CK_RV SoftHSM::C_GetMechanismInfo(CK_SLOT_ID slotID, CK_MECHANISM_TYPE type, CK_
...
+	AsymmetricAlgorithm* mldsa = CryptoFactory::i()->getAsymmetricAlgorithm(AsymAlgo::MLDSA);
+	if (mldsa != NULL)
+	{
+		mldsaMinSize = mldsa->getMinKeySize();
+		mldsaMaxSize = mldsa->getMaxKeySize();
+	}
+	else
+	{
+		return CKR_GENERAL_ERROR;
+	}
```
bukka:
```suggestion
	if (mldsa == NULL)
	{
		return CKR_GENERAL_ERROR;
	}
	mldsaMinSize = mldsa->getMinKeySize();
	mldsaMaxSize = mldsa->getMaxKeySize();
```

src/lib/SoftHSM.cpp
```diff
@@ -4421,6 +4471,58 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
...
+			mechanism = AsymMech::MLDSA;
+			bAllowMultiPartOp = false;
+			isMLDSA = true;
+			if (pMechanism->pParameter != NULL_PTR) {
+				if(pMechanism->ulParameterLen != sizeof(CK_SIGN_ADDITIONAL_CONTEXT))
+				{
+					ERROR_MSG("Invalid parameters");
+					return CKR_ARGUMENTS_BAD;
+				}
+				else
```
bukka: no need for `else`....

src/lib/SoftHSM.cpp
```diff
@@ -4421,6 +4471,58 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
...
 			isEDDSA = true;
 			break;
+#endif
+#ifdef WITH_ML_DSA
+		case CKM_ML_DSA:
+			mechanism = AsymMech::MLDSA;
+			bAllowMultiPartOp = false;
+			isMLDSA = true;
+			if (pMechanism->pParameter != NULL_PTR) {
+				if(pMechanism->ulParameterLen != sizeof(CK_SIGN_ADDITIONAL_CONTEXT))
```
bukka: NIT: space after if.

src/lib/SoftHSM.cpp
```diff
@@ -8151,11 +8386,11 @@ CK_RV SoftHSM::generateAES
 	if (rv == CKR_OK)
 	{
 		OSObject* osobject = (OSObject*)handleManager->getObject(*phKey);
-		if (osobject == NULL_PTR || !osobject->isValid()) 
+		if (osobject == NULL_PTR || !osobject->isValid())
```
bukka: please try to not add unrelated CS changes...

bukka (changes requested):
I did another round and managed to find just minor things. Except maybe that Session which I don't really like. I was checking it locally and don't see reason why it couldn't be done in the SoftHSM.cpp - might need extra allocation but think it's worth it not messing up the Session...

src/lib/SoftHSM.cpp
antoinelochet: I suspect VSCode to have done something on my back.
Anyway, I reverted all the CS changes that I have seen.

src/lib/session_mgr/Session.cpp
antoinelochet: I agree with you.
I have moved and refactored some of the logic already present in `AsymSignInit` and `AsynVerifyInit` and the logic added in `Session` in a static `MLDSAUtil::copyMLDSAContext` method.
Tell me what you think of it.

src/lib/SoftHSM.cpp
```diff
@@ -6354,23 +6522,23 @@ CK_RV SoftHSM::WrapKeySym
...
 			break;
-			
+
 		case CKM_DES3_CBC_PAD:
 			blocksize = 8;
 			wrappedlen = RFC5652Pad(keydata, blocksize);
 			algo = SymAlgo::DES3;
 			break;
-			
+
```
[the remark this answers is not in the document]
antoinelochet: ignore

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -0,0 +1,572 @@
...
+unsigned long OSSLMLDSA::getMinKeySize()
+{
+	return MLDSAParameters::ML_DSA_44_PRIV_LENGTH;
+}
+
+unsigned long OSSLMLDSA::getMaxKeySize()
+{
+	return MLDSAParameters::ML_DSA_87_PRIV_LENGTH;
+}
```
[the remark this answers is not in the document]
antoinelochet: ML-DSA key sizes are in bytes. See chapter 6.67.4 of PKCS#11 3.2

src/lib/SoftHSM.cpp
```diff
@@ -4155,6 +4198,10 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
 #endif
 #ifdef WITH_EDDSA
 	bool isEDDSA = false;
+#endif
+#ifdef WITH_ML_DSA
+	bool isMLDSA = false;
+	SIGN_ADDITIONAL_CONTEXT* additionalContext = NULL;
 #endif
```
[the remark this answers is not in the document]
antoinelochet: @coderabbitai Propose fix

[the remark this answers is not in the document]
antoinelochet: @coderabbiai replace the guard with a tiny helper in MLDSAUtil (e.g., freeMLDSAContext)

src/lib/SoftHSM.cpp
```diff
@@ -4155,6 +4198,11 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
...
 #ifdef WITH_EDDSA
 	bool isEDDSA = false;
+#endif
+#ifdef WITH_ML_DSA
+	bool isMLDSA = false;
+	// RAII guard to ensure no leak on early returns; null after setParameters().
+	MLDSAAdditionalContextGuard mldsaCtxGuard;
 #endif
```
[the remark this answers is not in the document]
antoinelochet: @coderabbitai Provide a fix

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -0,0 +1,557 @@
...
+	EVP_SIGNATURE_free(sig_alg);
+	if (verifyRV != 1) 
+	{
+        if (verifyRV != 0) 
+		{
+            ERROR_MSG("ML-DSA verify error (0x%08lX)", ERR_get_error());
+        }
+        return false;
+	}
+	return true;
```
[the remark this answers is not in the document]
antoinelochet: ignore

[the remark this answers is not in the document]
antoinelochet: @coderabbitai propose a fix for Memory leak in ML-DSA additional-context handling.

src/lib/crypto/MLDSAUtil.cpp
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	SIGN_ADDITIONAL_CONTEXT* sac = reinterpret_cast<SIGN_ADDITIONAL_CONTEXT*>(mem);
+	sac->hedgeType = hedgeType;
+	sac->contextLen = contextLength;
+	sac->contextData = (contextLength ? (static_cast<unsigned char*>(std::malloc(contextLength))) : NULL);
+	sac->ownsContextMemory = false; // memory is the single blob; freed by free(sac)
+	if (contextLength) {
+		std::memcpy(sac->contextData, contextAsChar, contextLength);
+	}
```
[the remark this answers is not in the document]
antoinelochet: @coderabbitai Proposed fix does not work.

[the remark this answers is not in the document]
antoinelochet: @coderabbitai it produces incorrect behavior where contextData value is incorrect when used in sign/verify operations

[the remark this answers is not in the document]
antoinelochet: @coderabbitai This is still not ok. Same problem, contextData is not the right value for sign/verify. It seems to be an issue  with pointer arithmetics.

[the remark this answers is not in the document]
antoinelochet: @coderabbitai Still not ok. The pointer arithmetic seems wrong because the contextData value in sign/verify is offset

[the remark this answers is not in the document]
antoinelochet: @coderabbitai This last solution ends up with a segmentation fault error.

src/lib/crypto/AsymmetricAlgorithm.h
```diff
@@ -116,6 +121,22 @@ struct RSA_PKCS_PSS_PARAMS
...
+		HEDGE_REQUIRED,
+		DETERMINISTIC_REQUIRED
+	};
+};
+
+struct SIGN_ADDITIONAL_CONTEXT
+{
+    Hedge::Type hedgeType;
+	ByteString* additionalContext;
+};
```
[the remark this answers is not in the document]
antoinelochet: This project is C++ 11, this suggestion does not work.

antoinelochet:
@coderabbitai review

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -225,39 +226,25 @@ bool OSSLMLDSA::verify(PublicKey *publicKey, const ByteString &originalData,
...
+		ERROR_MSG("Invalid mechanism parameter type supplied");
 
 		return false;
 	}
 
-	OSSL_PARAM params[4], *p = params;
-	SIGN_ADDITIONAL_CONTEXT* signAdditionalContext = (SIGN_ADDITIONAL_CONTEXT*) param;
-	ByteString* context;
+	OSSL_PARAM params[3], *p = params;
+	MLDSAMechanismParam* signAdditionalContext = (MLDSAMechanismParam*) mechanismParam;
```
bukka: It would be more cpp to use dynamic_cast here...

src/lib/crypto/AsymmetricAlgorithm.h
```diff
@@ -147,13 +132,13 @@ class AsymmetricAlgorithm
 	virtual ~AsymmetricAlgorithm() { }
 
 	// Signing functions
-	virtual bool sign(PrivateKey* privateKey, const ByteString& dataToSign, ByteString& signature, const AsymMech::Type mechanism, const void* param = NULL, const size_t paramLen = 0);
+	virtual bool sign(PrivateKey* privateKey, const ByteString& dataToSign, ByteString& signature, const AsymMech::Type mechanism, const void* param = NULL, const size_t paramLen = 0, const MechanismParam* mechanismParam = NULL);
```
bukka, line 135: Have you thought about incorporating param and paramLen into this new class so signature gets simlpler. I guess we can do it later but just wondering if it would make sense?

antoinelochet: We could but I think that it should left for another PR and make param and paramLen disappear altogether from the sign/verify methods signatures.
This PR has lingered and went through too much already :D

src/lib/crypto/OSSLMLDSA.cpp
antoinelochet: Done

src/lib/crypto/OSSLUtil.cpp
```diff
@@ -209,3 +221,16 @@ int OSSL::byteString2oid(const ByteString& byteString)
...
+
+	std::map<unsigned long, const char*>::const_iterator it = mldsaAlgNameFromParameterSet.find(parameterSet);
+
+	if (it != mldsaAlgNameFromParameterSet.end()) {
+		return it->second;
+	}
+
+	return NULL;
+}
+#endif
```
bukka: Missing new line

src/lib/session_mgr/Session.h
```diff
@@ -170,6 +175,7 @@ class Session
 
 	// Symmetric Crypto
 	SymmetricKey* symmetricKey;
+
```
bukka: NIT: accidentally added NL I guess...

bukka (commented):
Except few NIT's, it looks all good to me. I think we we can merge it as soon as it gets sorted. If you could squash it and add some nice description to the commit , that would be great!

src/lib/crypto/OSSLUtil.cpp
antoinelochet: Fixed

src/lib/session_mgr/Session.h
antoinelochet: Fixed

PR #816, opened 2025-09-11
Support RSA PSS for import in softhsm2-util

### LINKED ISSUE #721: Import fails with RSA-PSS keys
Test:

```
$ softhsm2-util --version
2.6.1

$ openssl version
OpenSSL 3.1.2 1 Aug 2023 (Library: OpenSSL 3.1.2 1 Aug 2023)

$ softhsm2-util --init-token --free --label rsapsstest --so-pin 1234 --pin 1234
Slot 1 has a free/uninitialized token.
The token has been initialized and is reassigned to slot 456343862

$ openssl genpkey -algorithm RSA-PSS -pkeyopt rsa_keygen_bits:2048 -out rsa2048.pkcs8.pem

$ softhsm2-util --so-pin 1234 --pin 1234 --token rsapsstest --import rsa2048.pkcs8.pem --label rsa2048 --id 0001
Found slot 456343862 with matching token label.
ERROR: Cannot handle this algorithm.
```

bukka:
This should fix #721 . The type is differnt from RSA PSS so it needs to be added there.

I tested the change manual and the import went through after this. I don't see any tests for util so it comes wihtout a test.

## Summary by CodeRabbit

- **New Features**
  - Added support for importing RSA-PSS key pairs; RSA-PSS keys are now recognized and processed the same as standard RSA during import.
  - Existing import behavior for RSA, DSA, EC, and EdDSA remains unchanged.
  - No changes to public interfaces, CLI options, or configuration are required.

### DIFF
--- a/src/bin/util/softhsm2-util-ossl.cpp
+++ b/src/bin/util/softhsm2-util-ossl.cpp
@@ -156,6 +156,7 @@ int crypto_import_key_pair
 	switch (EVP_PKEY_type(EVP_PKEY_id(pkey)))
 	{
 		case EVP_PKEY_RSA:
+		case EVP_PKEY_RSA_PSS:
 			rsa = EVP_PKEY_get1_RSA(pkey);
 			break;
 		case EVP_PKEY_DSA:


bukka (commented):
I'm going to merge this as there is really not much to review and it fixes the mentioned issue

PR #828, opened 2025-12-06
Fix ECC OpenSSL cmake check on Windows

bukka:
There is currently failure in vcpkg build so this PR is an attempt to try to fix it.

## Summary by CodeRabbit

* **Chores**
  * Enhanced Windows build setup for the OpenSSL backend: the build now augments the process PATH with the matching OpenSSL/vcpkg runtime bin directory when available, enabling ECC-related tests to run on Windows without other logic changes.


bukka:
@jschlyter Somehow managed to get Windows (and everything else) green. Windows Botan is a bit slow but not sure how could this be impacted by this change that is just for OpenSSL. So it should be hopefully safe.

PR #831, opened 2025-12-10
Fixed PKCS#11 3.2 C_Decapsulate definition

antoinelochet:
There is a slight difference between reference PKCS#11 3.2 functions header and the one from p11kit merged in #827
Definition of `CK_FLAGS_PTR` was also missing

## Summary by CodeRabbit

* **Refactor**
  * Updated cryptographic operation API parameters for improved type consistency.
  * Added new type definition aliases for flag parameters.
  * Standardized type declarations in compatibility layer.

*Note: These are API-level changes that may require updates for developers using this library.*


bukka (approved):
Ah looks like this is wrong in p11kit as I just copied it from there. Might be good to also report it to them so next time when we copy new header, we don't have the same issue.

PR #832, opened 2025-12-16
fix: object_store: stop throwing errors when checking generation file

### LINKED ISSUE #753: A lot of syslog messages for 'generation' file.
A lot of syslog messages are generated for 'generation' file.

Could not open the file (No such file or directory): /usr/tsk/tokens/ee5f7cbb-882e-7aaa-c648-12462c76cee7/generation

It seems the generation file is never created but for few operations it is tried to be open. The open request leads to syslog message flooding. This seems to be a known issue in softHSM but I am not sure if it is already fixed or there is a workaround.

Thanks.

amananas:
A new routine is added to File class, to check whether a file exists or not. It will allow file existence check, without trying to open the file and thus throw an error when the use case is valid.

This new routine is added to Generation wasUpdated routine: if the generation file does not exist, the routine will exit without error, instead of trying to open the file and report an error.

Fixes #753

## Summary by CodeRabbit

* **New Features**
  * Added a file-existence check so callers can determine whether a file exists without opening it.

* **Bug Fixes**
  * Update detection now treats missing files as needing updates, improving change-detection correctness.

* **Tests**
  * Expanded tests to validate file-existence checks before performing file operations.

### DIFF
--- a/src/lib/object_store/File.cpp
+++ b/src/lib/object_store/File.cpp
@@ -143,6 +143,20 @@ File::~File()
 	}
 }
 
+// Check if a file exists without trying to open it
+// May be used when a routine just wants to check file existence,
+// and does not want to throw an error by trying to open it.
+bool File::exists(const std::string& path)
+{
+#ifndef _WIN32
+	struct stat buf;
+	return stat(path.c_str(), &buf) == 0;
+#else
+	return _access(path.c_str(), 0) == 0;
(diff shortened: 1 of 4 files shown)


bukka (changes requested):
It makes sense but please also add test to FileTests.cpp

bukka (approved):

PR #850, opened 2026-03-19
Modify CKR_ATTRIBUTE_READ_ONLY Error when generating ML-DSA key pair using pkcs11-tool

bgkim-ictk:
Modify CKR_ATTRIBUTE_READ_ONLY Error when generating ML-DSA key pair using following pkcs11-tool command

* pkcs11-tool --module /usr/local/lib/softhsm/libsofthsm2.so --slot 1944964809 --login --pin 1234 --keypairgen --key-type ML-DSA-44 --id 0  --label "ML-DSA-44"

please review

## Summary by CodeRabbit

* **New Features**
  * ML-DSA support is now configurable across supported cryptographic backends.

* **Bug Fixes**
  * Refined key generation to properly handle parameter set attributes during object creation operations.

### DIFF
--- a/cmake/modules/CompilerOptions.cmake
+++ b/cmake/modules/CompilerOptions.cmake
@@ -219,6 +219,10 @@ if(WITH_CRYPTO_BACKEND STREQUAL "botan")
         message(STATUS "Botan: Support for EDDSA is disabled")
     endif(ENABLE_EDDSA)
 
+    if(ENABLE_MLDSA)
+        set(WITH_ML_DSA 1)
+    endif(ENABLE_MLDSA)
+
     # acx_botan_gost.m4
     if(ENABLE_GOST)
         set(testfile ${CMAKE_SOURCE_DIR}/cmake/modules/tests/test_botan_gost.cpp)
@@ -369,6 +373,10 @@ elseif(WITH_CRYPTO_BACKEND STREQUAL "openssl")
     else(ENABLE_EDDSA)
         message(STATUS "OpenSSL: Support for EDDSA is disabled")
     endif(ENABLE_EDDSA)
+
+    if(ENABLE_MLDSA)
+        set(WITH_ML_DSA 1)
+    endif(ENABLE_MLDSA)
 
     # acx_openssl_gost.m4
     if(ENABLE_GOST)
--- a/src/lib/P11Attributes.cpp
+++ b/src/lib/P11Attributes.cpp
@@ -440,8 +440,11 @@ CK_RV P11Attribute::update(Token* token, bool isPrivate, CK_VOID_PTR pValue, CK_
 	{
 		if (OBJECT_OP_GENERATE==op)
 		{
-			ERROR_MSG("Prohibited attribute was passed to key generation function");
-			return CKR_ATTRIBUTE_READ_ONLY;
+            if (type != CKA_PARAMETER_SET) 
+            {
+                ERROR_MSG("Prohibited attribute was passed to key generation function");
+                return CKR_ATTRIBUTE_READ_ONLY;
+            }
 		}
 	}
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,6 +122,8 @@ enable_testing()
 
 # config.h include path
 include_directories(${CMAKE_BINARY_DIR})
+include_directories(${CMAKE_SOURCE_DIR}/src/lib/slot_mgr)
+include_directories(${CMAKE_SOURCE_DIR}/src/lib/object_store)
 
 add_subdirectory(src)
 


cmake/modules/CompilerOptions.cmake
```diff
@@ -219,6 +219,10 @@ if(WITH_CRYPTO_BACKEND STREQUAL "botan")
         message(STATUS "Botan: Support for EDDSA is disabled")
     endif(ENABLE_EDDSA)
 
+    if(ENABLE_MLDSA)
+        set(WITH_ML_DSA 1)
+    endif(ENABLE_MLDSA)
```
bukka, line 224: MLDSA currently does not work with Botan so this should be removed

src/lib/P11Attributes.cpp
```diff
@@ -440,8 +440,11 @@ CK_RV P11Attribute::update(Token* token, bool isPrivate, CK_VOID_PTR pValue, CK_
 	{
 		if (OBJECT_OP_GENERATE==op)
 		{
-			ERROR_MSG("Prohibited attribute was passed to key generation function");
-			return CKR_ATTRIBUTE_READ_ONLY;
+            if (type != CKA_PARAMETER_SET)
```
bukka, line 443: This seems a bit hacky. I might need to debug it to see what the actual reason for this.

antoinelochet:
> CC @antoinelochet for review

I did not try to generate ML-DSA keys with anything but calling the library directly.
I do not know what pkcs11-tool does in this case.
Like you, I will have to checkout the branch and see what is happening.

cmake/modules/CompilerOptions.cmake
antoinelochet: We would need Botan3 for that which is not currently supported. It will fail with Botan2.

antoinelochet:
I have checked with the PKCS#11 3.2 spec (https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/csd01/pkcs11-spec-v3.2-csd01.html#_Toc195693264) and the OpenSC/pkcs11-tool 0.27.1 implementation.
If my interpretation is correct, here are my findings.

While the spec says that ML-DSA private key should not set the CKA_PARAMETER_SET attribute on the C_GenerateKeyPair operation (Chapter 6.67.3 with Table 13, note 4), pkcs11-tool does it: https://github.com/OpenSC/OpenSC/blob/19868984dc4dc697af6a86d65ab32a1f19a43ea4/src/tools/pkcs11-tool.c#L3730
The SoftHSM implementation of CKA_PARAMETER_SET is respectful of the PKCS#11 3.2 specs and forbids setting CKA_PARAMETER_SET on the private key template through the use of P11Attribute::ck4.
So it seems an issue on pkcs11-tool side. Maybe it has been done because other PKCS#11 implementations do not care. I don't know.
I see several options in my order of preference:
- We can stay as-is because we want to stick as close as possible to the spec.
- We can check the presence of the attribute on the private template on the C_GenerateKeyPair call and remove it (while logging this problem in DEBUG level) so that we do not change the current implementation.
- We can remove the P11Attribute::ck4 constraint and add a coherence check between the public and private templates.

@bgkim-ictk @bukka what do you think ?

bukka:
It would be great to know what was the reasoning for it in pkcs11-tool. Maybe @Jakuje knows it.

Jakuje:
I think the SHOULD is a recommendation. Hard rejecting the key generation when private key template contains the parameter set does not sound like a good idea (as long as it is matching the parameter set on public key).

I do not think this was an intention, but when I was writing that, there was not many pkcs11 tokens with this in the wild. I think it was more likely an overlook which was not noticed by anyone before. Please open a PR/issue on OpenSC to fix that.

antoinelochet:
On the Table 13, the notes are 'MUST': "4 MUST not be specified when object is generated with C_GenerateKey or C_GenerateKeyPair."

Jakuje:
> On the Table 13, the notes are 'MUST': "4 MUST not be specified when object is generated with C_GenerateKey or C_GenerateKeyPair."

Ok. My bad then. Its then clearly a bug in pkcs11-tool to be fixed. Though accepting it should not hurt on the sofhsm side, unless you aim to be bug-bug-compatible with pkcs11 specs

antoinelochet:
I agree. I did not test setting CKA_PARAMETER_SET on privat ekey template any other tokens that I have on hand.
I may do that next week.
In the meantime, what do you think of a solution where first we check that CKA_PARAMETER_SET value is the same on the public and private template. Second, remove it from the private template and trace this removal through a log message.

Jakuje:
> I agree. I did not test setting CKA_PARAMETER_SET on privat ekey template any other tokens that I have on hand.

I was testing this mostly against kryoptic, which was the only token I was aware having these mechanism at that time. And we are clearly not that strict in this regard.

>  In the meantime, what do you think of a solution where first we check that CKA_PARAMETER_SET value is the same on the public and private template. Second, remove it from the private template and trace this removal through a log message.

I am not familiar with the softhsm internals of handling of the attributes, but I think this should work.

antoinelochet:
I did some testing on Thales Luna7 and Bull Proteccio. Luna7 does not care the CKA_PARAMETER_SET is set on private key template, Proteccio does.
I suggest that we stick to my latest proposed solution.
@bgkim-ictk @bukka is that ok for you ?

Jakuje:
For the record, I opened https://github.com/OpenSC/OpenSC/pull/3640 to fix this. But it will get some time before it will be in a release so I think implementing a workaround in softhsm would still make sense.

PR #851, opened 2026-03-21
Optimize handles clearing on session / token closing

bukka:
Currently the full map scan is done on each clearing. This can result in significant CPU usage when large number of objects are used. The main problem is that full scan is done when any session is closed so if application has large number in one session, they get scanned even though there is no need for that.

This PR introduces additional maps that track object per session / slot as well as slot session count.

## Summary by CodeRabbit

* **Refactor**
  * Improved performance of handle manager operations by optimizing internal data structures to reduce lookup times during session and slot cleanup operations.


antoinelochet:
Very nice !
I had some issues in the past because of high number of objects.

bukka:
Thanks for checking. I let my client test this quite a bit and no issue visible there so merging.

PR #854, opened 2026-03-23
fix: check key type in C_SignInit/C_VerifyInit to prevent crash

mingulov:
Passing a wrong asymmetric key type (e.g. EC key) to a mismatched mechanism (for example RSA related) caused a crash (segfault or abort) because the code tried to read RSA attributes from an EC key object and passed invalid data to OpenSSL.

Add key type validation after the mechanism switch in both functions, returning CKR_KEY_TYPE_INCONSISTENT on mismatch. Also convert the GOST catch-all else branch to an explicit else-if(isGOST) check.

Regression tests use cross-matched asymmetric keys (EC key with RSA mechanisms, RSA key with ECDSA/EdDSA mechanisms) to verify the fix. Without the fix, C_SignInit accepts the wrong key type and the subsequent C_Sign crashes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened key-type validation for sign/verify initialization, returning clear errors when a key and mechanism mismatch.
  * Clarified handling for GOST and ML-DSA mechanisms to ensure proper mechanism validation.

* **Tests**
  * Added tests covering key-type mismatch scenarios for sign and verify initialization.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -4190,6 +4190,9 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
 	if (!isMechanismPermitted(key, pMechanism->mechanism))
 		return CKR_MECHANISM_INVALID;
 
+	// Get key info
+	CK_KEY_TYPE keyType = key->getUnsignedLongValue(CKA_KEY_TYPE, CKK_VENDOR_DEFINED);
+
 	// Get the asymmetric algorithm matching the mechanism
 	AsymMech::Type mechanism = AsymMech::Unknown;
 	void* param = NULL;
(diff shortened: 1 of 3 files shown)


src/lib/SoftHSM.cpp
```diff
@@ -4527,6 +4532,24 @@ CK_RV SoftHSM::AsymSignInit(CK_SESSION_HANDLE hSession, CK_MECHANISM_PTR pMechan
...
+		return CKR_KEY_TYPE_INCONSISTENT;
+#endif
+#ifdef WITH_GOST
+	if (isGOST && keyType != CKK_GOSTR3410)
+		return CKR_KEY_TYPE_INCONSISTENT;
+#endif
+#ifdef WITH_EDDSA
+	if (isEDDSA && keyType != CKK_EC_EDWARDS)
+		return CKR_KEY_TYPE_INCONSISTENT;
+#endif
```
bukka: It would be a bit better to do those check in the blocks below and introduce new onews for isEDDSA and isECDSA

mingulov: Done, updated.
Check moved to block below, for the consistency - ML-DSA key type check is also moved.

PR #858, opened 2026-04-14
Fix memory leak in EdDSA PrintableString handling

olszomal:
Fix a small memory leak in `OSSL::byteString2oid()` when CKA_EC_PARAMS is decoded as ASN1_PRINTABLESTRING for EdDSA keys.

Also replace `strcmp()` on ASN.1 string data with length-aware `memcmp()` using `ASN1_STRING_get0_data()` and `ASN1_STRING_length()`, as ASN.1 strings are not guaranteed to be NUL-terminated.

For consistency, also free the temporary ASN1_OBJECT in the OID path.

Detected via Valgrind (ASN1_PRINTABLESTRING leak during C_SignInit with EdDSA keys).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing and validation for Edwards25519/Curve25519 and Edwards448/Curve448 curve names.
  * Added null-safety and exact-length checks to prevent misidentification.
  * Ensured proper cleanup of decoded objects to improve stability and memory safety.
  * No public API changes.

### DIFF
--- a/src/lib/crypto/OSSLUtil.cpp
+++ b/src/lib/crypto/OSSLUtil.cpp
@@ -172,12 +172,14 @@ ByteString OSSL::oid2ByteString(int nid)
 // Convert a ByteString to an OpenSSL EVP_PKEY id
 int OSSL::byteString2oid(const ByteString& byteString)
 {
-	ASN1_OBJECT *oid = NULL;
-	ASN1_PRINTABLESTRING *curve_name = NULL;
+	ASN1_OBJECT *oid;
+	ASN1_PRINTABLESTRING *curve_name;
 	const unsigned char *p = byteString.const_byte_str();
 	const unsigned char *pp = p;
+	const unsigned char *data;
 	long length;
-	int tag, pclass;
+	int tag, pclass, data_len;
+	int nid = NID_undef;
 
 	ASN1_get_object(&pp, &length, &tag, &pclass, byteString.size());
 	if (pclass == V_ASN1_UNIVERSAL && tag == V_ASN1_OBJECT)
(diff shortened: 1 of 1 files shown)


src/lib/crypto/OSSLUtil.cpp
```diff
@@ -190,35 +192,46 @@ int OSSL::byteString2oid(const ByteString& byteString)
...
-			return EVP_PKEY_ED25519;
+			return NID_undef;
+		}
+
+		data = ASN1_STRING_get0_data(curve_name);
+		data_len = ASN1_STRING_length(curve_name);
+
+		if (data_len == 12 && memcmp(data, "edwards25519", data_len) == 0)
+		{
+			nid = EVP_PKEY_ED25519;
```
bukka:
```suggestion
			ASN1_PRINTABLESTRING_free(curve_name);
			return EVP_PKEY_ED25519;
```

bukka: This is so it skips other checks...

bukka (commented):
Good catch of the mem leak! Just a minor suggestion of not repeating those data_len checks which I'm not 100% sure every compiler optimizes out...

bukka:
Oh sorry I was just going through PR in order from the latest and merged https://github.com/softhsm/SoftHSMv2/pull/863 . I was thinking that it is familiar but didn't realise that you fixed the same thing. This actually still has got that curve check and slightly more optimal comparison so if you could strip what got merged and keep what's left, that would be awesome.

olszomal:
No worries.  I rebased onto #863, removed the already merged parts, and kept the remaining improvements. Should be clean now.

bukka (approved):

PR #864, opened 2026-04-25
use global flag for initialised state

### LINKED ISSUE #729: openssl operations involving pcks11 and softHSM result in segfault on exit

### LINKED ISSUE #780: softhsm2 crashes in context of openssl 3 pkcs11 provider
Command like this one
```sh
openssl pkeyutl -sign $provider -inkey "pkcs11:id=$key_id" -in $data -out $signature  -passin "pass:$userpin"
```
dumps core  on exit.

Stack trace shows:
```
Core was generated by `.../bin/openssl pkeyutl -sign -inkey pkcs11:id=%0'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f549cdbd0ae in HandleManager::getSession(unsigned long) () from .../lib64/softhsm/libsofthsm2.so
(gdb) bt
#0  0x00007f549cdbd0ae in HandleManager::getSession(unsigned long) () from .../lib64/softhsm/libsofthsm2.so
#1  0x00007f549cd6b7cf in SoftHSM::C_CloseSession(unsigned long) () from .../lib64/softhsm/libsofthsm2.so
#2  0x00007f549cd51e44 in C_CloseSession () from .../lib64/softhsm/libsofthsm2.so
#3  0x00007f549d3ef442 in p11prov_CloseSession (ctx=0x33e17390, hSession=1) at .../pkcs11-provider-current/src/interface.gen.c:283
#4  0x00007f549d40515f in token_session_close (session=session@entry=0x33e42070) at .../pkcs11-provider-current/src/session.c:110
#5  0x00007f549d4052e0 in token_session_close (session=0x33e42070) at .../pkcs11-provider-current/src/session.c:358
#6  session_free (session=0x33e42070) at .../pkcs11-provider-current/src/session.c:382
#7  0x00007f549d405dbd in p11prov_session_pool_free (pool=0x33e41380) at .../pkcs11-provider-current/src/session.c:170
#8  0x00007f549d40bb04 in p11prov_free_slots (sctx=0x33e35d50) at .../pkcs11-provider-current/src/slot.c:354
#9  0x00007f549d402530 in p11prov_ctx_free (ctx=0x33e17390) at .../pkcs11-provider-current/src/provider.c:545
#10 0x00007f549d05d1b5 in ossl_provider_free (prov=0x33e15f70) at crypto/provider_core.c:720
#11 0x00007f549d183161 in OPENSSL_sk_pop_free (st=0x33df0af0, func=0x7f549d05d220 <provider_deactivate_free>) at crypto/stack/stack.c:439
#12 0x00007f549d05bc69 in sk_OSSL_PROVIDER_pop_free (freefunc=0x7f549d05d220 <provider_deactivate_free>, sk=<optimized out>) at crypto/provider_core.c:198
#13 ossl_provider_store_free (vstore=0x33df0a90) at crypto/provider_core.c:294
#14 0x00007f549d045e45 in context_deinit_objs (ctx=0x7f549d37e580 <default_context_int>) at crypto/context.c:271
#15 context_deinit (ctx=0x7f549d37e580 <default_context_int>) at crypto/context.c:355
#16 ossl_lib_ctx_default_deinit () at crypto/context.c:396
#17 0x00007f549d04a615 in OPENSSL_cleanup () at crypto/init.c:440
#18 0x00007f549ca5a1a6 in __run_exit_handlers () from /lib64/libc.so.6
#19 0x00007f549ca5a2ec in exit () from /lib64/libc.so.6
#20 0x000000000042c0dd in main (argc=<optimized out>, argv=<optimized out>) at apps/openssl.c:320
```

---- 
Question is how SoftHSMv2 manages "global" variables?

petrovr:
Resolves https://github.com/softhsm/SoftHSMv2/issues/780 , and closes https://github.com/softhsm/SoftHSMv2/issues/729 .

## Summary by CodeRabbit

* **New Features**
  * Added a PKCS#11/OpenSSL utility to initialize software tokens, generate/import keys, and perform signing and verification when OpenSSL support is available.

* **Chores**
  * Build/test configuration updated to include and conditionally run the new utility.
  * VCS ignore and cleanup rules added to exclude and remove local token data.

* **Refactor**
  * Token initialization state made global to ensure consistent behavior across the component lifecycle.


petrovr:
Pushed updated patch that addresses all nits.

petrovr:
Pushed updated patch improve use or remove of quotation marks.

petrovr:
Pushed updated patch to address https://github.com/softhsm/SoftHSMv2/pull/864#pullrequestreview-4176957511

petrovr:
address https://github.com/softhsm/SoftHSMv2/pull/864#pullrequestreview-4176988287 , but some shell variables will remain unquoted as value is without space

src/bin/util/test/p11prov-test.sh
```diff
@@ -0,0 +1,160 @@
+#! /bin/sh
```
bukka, line 1: Might be probably better to add it to to test subdit (it means src/bin/util/test) so it's clear it's a test.

bukka (commented):
It looks good from a quick look and static seems to make more sense. Will check it again later.

petrovr:
Test has been moved to src/bin/test, similar to src/lib/test .

petrovr:
I note that now exist "test" directory for utility. Test moved to it. 

---- 

Off topic : ML-KEM and ML-DSA tests fail if build is with OpenSSL < 3.5.

antoinelochet:
> I note that now exist "test" directory for utility. Test moved to it.
> 
> Off topic : ML-KEM and ML-DSA tests fail if build is with OpenSSL < 3.5.

This should have been fixed since your comment.

bukka:
There are some Windows failure. Could you please rebase this?

src/bin/util/test/p11prov-test.sh
```diff
@@ -0,0 +1,160 @@
...
+
+softhsm2_tool() {
+"$CWD"/../softhsm2-util --module "$P11MODULE" ${1+"$@"}
+}
+
+
+# configurations
+TOKEN_DIR="$CWD"/tokens
+rm -rf "$TOKEN_DIR"
+mkdir "$TOKEN_DIR"
```
bukka, line 131: This should clean the tokens after the test in the same way as `src/bin/util/test/import-key-test-common.sh` does

src/bin/util/test/p11prov-test.sh
```diff
@@ -0,0 +1,160 @@
...
+
+openssl_version=`openssl version` || exit $?
+if test -z "$openssl_version" ; then
+  echo "cannot determine OpenSSL version" >&2
+  exit 99
+fi
+
+case $openssl_version in
+*"OpenSSL 0.9."*|\
+*"OpenSSL 1."*)
```
bukka, line 27: There should be also skip for LibreSSL.

src/bin/util/test/p11prov-test.sh
```diff
@@ -0,0 +1,160 @@
...
+  echo "$openssl_version is not impacted" >&2
+  exit 77
+  ;;
+esac
+# NOTE OpenSSL > 1.*
+
+
+# find a PKCS#11 provider
+p11_find_provider() {
```
bukka, line 43: I suppose this doesn't work in the CI as there is no pkcs11-provider installed - it might be good to extend CI for that so it's not purely local test.

src/bin/util/test/p11prov-test.sh
bukka: hmm it might just get covered by the provider test but wouldn't hurt to add it...

src/bin/util/test/.gitignore
```diff
@@ -0,0 +1 @@
+/tokens
```
bukka, line 1: hmm I don't think we need this if there is a proper clean up as suggested below.

bukka (commented):
It looks good. Just some minor comments on the test but the main logic is sane and should fix that issue.

I noticed that we should probably also use it before calling detechFork as that forkId might have the similar issue. So just doing something like
```
else if (isInitialised && instance->detectFork())
```
should do.

It would be probably good to also add it to CMakeList with skip 77 to make clear that it's not currently supported there but it's a good reminder that it could be added for Windows later... I'm not asking you to such support here so it's not too much extra work for you but eventually we would like to run it on Windows...

src/bin/util/test/p11prov-test.sh
petrovr: > There should be also skip for LibreSSL.

fixed by new commit "restrict tests to OpenSSL 3+: address"

petrovr:
Please note that I prefer to marge new commits manually.

src/bin/util/test/p11prov-test.sh
petrovr: > I suppose this doesn't work in the CI as there is no pkcs11-provider installed - it might be good to extend CI for that so it's not purely local test.

This is managed by exist status - 77 means skip tests.
Also if is installed the test should pass..

src/bin/util/test/p11prov-test.sh
petrovr: > This should clean the tokens after the test in the same way as `src/bin/util/test/import-key-test-common.sh` does

Fixed by new commit "clean-up "token" directory on success"

src/bin/util/test/.gitignore
petrovr: > hmm I don't think we need this if there is a proper clean up as suggested below.

Please let keep this.
For instance user my cancel unfinished test and git should ignore tests artifacts.

petrovr:
> It looks good. Just some minor comments on the test but the main logic is sane and should fix that issue.
> 
> I noticed that we should probably also use it before calling detechFork as that forkId might have the similar issue. So just doing something like
> 
> ```
> else if (isInitialised && instance->detectFork())
> ```
> 
> should do.
> 
> It would be probably good to also add it to CMakeList with skip 77 to make clear that it's not currently supported there but it's a good reminder that it could be added for Windows later... I'm not asking you to such support here so it's not too much extra work for you but eventually we would like to run it on Windows...

Partiality is fixed by new commit "add test to CMake rules".
Also the test scripts uses new model to find pksc#11 module. For Instance CMake build system may use different subdirectory depending from build type.

petrovr:
One of new commit is:  use "-test.sh" suffix.
The substring "-test. is not important. It just follow naming convention for other commits.

Key is ".sh" suffix.  Let discuss this part in context of Microsoft Windows build.

The CMake tests run bourne shell script using bash program. My understanding is that bash accept only Unix EOL. The native GIT command extract .sh as text files i.e., with  Microsoft Windows end of line model - <CR><LF>.
Expected is test to fail.

This could be resolved with explicit declaration. Foe example - https://gitlab.com/termapps/termoneplus/-/blob/master/.gitattributes?ref_type=heads .

The build system is run on two OS-es - Linux or Microsoft Windows. Important is :
```
# shell files have unix(LF) end-of-line conversion

*.sh text eol=lf
```

----

This is reason to add .sh suffix .and my question is about creation of ".gitattributes". 

Can I add rules to this pull request, or do I need to create a separate issue?

src/bin/util/test/p11prov-test.sh
bukka: What I meant is that it might be useful to install pkcs11-provider in CI so we have got at least one workflow where this test runs. In other words I would like to have this running in CI and not be always skipped. The test that runs only in local setup gets much more easily broken without anyone noticing.

src/bin/util/test/.gitignore
bukka: Ideally we would prefer to have something that always cleans it up but in the current setup (bash script direct execution), it's not exactly easy so I don't have issue with this being kept.

src/bin/util/test/p11prov-test.sh
petrovr: done with a separate commit - add pkcs11-provider to the CI workflow tests with host OpenSSL 3+

petrovr:
> Please note that I prefer to marge new commits manually.

Previous set merged (manual fixup) to one commit.

The new set adds commit "get path to SoftHSM utility". I cannot test on come OS-es. I try to simulate "multi target build" on Linux with cmake ninja build. The improvement is in this commit and I guess that this address build with other systems.

petrovr:
Updated to only use shell glob pattern to find file and search modifications merged into main commit.

PR #867, opened 2026-04-28
Added PKCS#11 3.2 ML-DSA multiple-part support in lib

antoinelochet:
Adds the multiple-part support for ML-DSA.
Follow-up of #809 

## Summary by CodeRabbit

* **New Features**
  * Enabled multipart ML‑DSA signing and verification. Init APIs accept mechanism-specific parameters and now log ML‑DSA context/hedge details when provided.

* **Tests**
  * Added multipart ML‑DSA tests for streaming sign/verify flows, including a variant exercising additional context/hedge parameters.

### DIFF
--- a/src/lib/crypto/OSSLMLDSA.cpp
+++ b/src/lib/crypto/OSSLMLDSA.cpp
@@ -30,6 +30,7 @@ bool OSSLMLDSA::sign(PrivateKey *privateKey, const ByteString &dataToSign,
 					 const void * /* param  = NULL*/, const size_t  /* paramLen = 0 */,
 					 const MechanismParam* mechanismParam)
 {
+	DEBUG_MSG("sign dataToSign: %s", dataToSign.hex_str().c_str());
 	if (mechanism != AsymMech::MLDSA)
 	{
 		ERROR_MSG("Invalid mechanism supplied (%i)", mechanism);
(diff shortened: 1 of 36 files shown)


src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -30,6 +30,7 @@ bool OSSLMLDSA::sign(PrivateKey *privateKey, const ByteString &dataToSign,
 					 const void * /* param  = NULL*/, const size_t  /* paramLen = 0 */,
 					 const MechanismParam* mechanismParam)
 {
+	DEBUG_MSG("sign dataToSign: %s", dataToSign.hex_str().c_str());
```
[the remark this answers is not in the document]
antoinelochet: Ignore

src/lib/crypto/OSSLMLDSA.cpp
```diff
@@ -155,26 +156,82 @@ bool OSSLMLDSA::sign(PrivateKey *privateKey, const ByteString &dataToSign,
...
-	ERROR_MSG("ML-DSA does not support multi part signing");
+	if (!AsymmetricAlgorithm::signUpdate(dataToSign))
+	{
+		return false;
+	}
 
-	return false;
+	DEBUG_MSG("signUpdate dataToSign %s", dataToSign.hex_str().c_str());
+
+	message += dataToSign;
```
bukka, line 211: hmm this is somehow killing the reason to do the multipart operation and might result in significant memory increase for long messages.

I realise that it is the same situation as EdDSA where len is needed in advance.

So the question is if we really need to support multipart for those algs?

antoinelochet: OpenSSL does not support multi-part signature. At least for ML-DSA.
So, as you have seen, I have copied the way EdDSA multi-part is implemented.
On your reflections:
- I don't necessarily think that long message are that significant. In cases where messages get really long and a hardware token, it should be preferable to use Pre-Hash ML-DSA. Which we could implement using the trick described in OpenSSL ML-DSA design document: https://github.com/openssl/openssl/blob/master/doc/designs/ml-dsa.md
- I think that the support is worth adding for completness sake. I know that the Thales Luna7 is able to do multi-part sign/verify on ML-DSA so it is nice to have a comparable feature set with "real" tokens.

bukka: Ok if some devices supports it, then it probably makes sense to support it here as well.

bukka (approved):

Issue #856, opened 2026-03-26
CBC supports wrap but not unwrap


Leont:
`CKM_AES_CBC` (and `CKM_DES3_CBC`) seem to support the wrap operation, but not the unwrap operation, this is kind of silly.

Given that `AES_CBC_PAD` does implement both, I suspect it's easy to add unwrapping support (also it looks like the unpadded wrapping code needs to be fixed to actually copy the IV).

Hussainity:
I have a need for this as well, happy to take up implementation if no one is working on this.

PR #871, opened 2026-05-01
Added support for AES CBC Key Unwrap

Hussainity:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AES-CBC block/IV initialization and handling during key wrap, ensuring correct IV length and no unintended unpadding for non-padding AES-CBC.
  * Updated mechanism reporting so AES-CBC is advertised as supporting both wrap and unwrap; added IV-length validation for AES-CBC unwrap.

* **Tests**
  * Added tests for AES-CBC (non-padding) wrap/unwrap and adjusted wrap-length expectations.

### DIFF
--- a/src/lib/SoftHSM.cpp
+++ b/src/lib/SoftHSM.cpp
@@ -6627,7 +6627,6 @@ CK_RV SoftHSM::WrapKeySym
 			break;
 
 		case CKM_DES3_CBC:
-			blocksize = 8;
 			algo = SymAlgo::DES3;
 			break;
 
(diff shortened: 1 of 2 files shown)


Hussainity:
Resolves the CBC Unwrap portion of this issue: #856

src/lib/SoftHSM.cpp
```diff
@@ -6626,6 +6627,7 @@ CK_RV SoftHSM::WrapKeySym
 			break;
 
 		case CKM_DES3_CBC:
```
antoinelochet, line 6629: There is PR https://github.com/softhsm/SoftHSMv2/pull/812 about CKM_DES3_CBC. I think it would be better to remove this change.

antoinelochet:
Apart from my comment about CKM_DES3_CBC blocksize, this seems good !

Leont:
Looks sensible to me

bukka (changes requested):
I agree with @antoinelochet comment so please get rid of that bit and then it should be fine to merge.

src/lib/SoftHSM.cpp
Hussainity: Addressed

Hussainity:
Thanks all for the review!

Hussainity:
@bukka requesting review

bukka (approved):

PR #875, opened 2026-05-17
Fix leak of session objects on session close

bukka:
SessionObjects were kept alive in allObjects as tombstones with valid=false after their session closed, so every object ever created lingered until C_Finalize, holding its Mutex*. Long-running clients saw unbounded memory growth dominated by mutex objects.

Add refcounting to OSObject and have SessionObjectStore, HandleManager, and C_FindObjectsInit acquire/release across the boundaries where session-object pointers escape the store mutex. Releases happen outside the relevant lock to avoid widening critical sections. Also erase from allObjects in the store cleanup paths to avoid a double-free against the destructor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread-safety and stability around object lifetime and session closure to prevent use-after-free and avoid releasing objects while locks are held.
  * More reliable object enumeration during searches and session/token operations for consistent results.

* **Refactor**
  * Introduced explicit acquire/release ownership semantics and updated handle management to use typed references for safer cleanup and reduced races.

* **Tests**
  * Updated tests to validate presence, validity, and lifecycle behavior under the new ownership model.

[Review Change Stack](https://app.coderabbit.ai/change-stack/softhsm/SoftHSMv2/pull/875?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/handle_mgr/HandleManager.cpp
+++ b/src/lib/handle_mgr/HandleManager.cpp
@@ -56,40 +56,52 @@ HandleManager::HandleManager()
 // Destructor
 HandleManager::~HandleManager()
 {
+	// Safety net to release any session-object handles that weren't explicitly cleaned up.
+	// In normal C_Finalize flow this should be empty.
+	for (auto& kv : handles)
+	{
+		if (kv.second.kind == CKH_OBJECT && kv.second.hSession != CK_INVALID_HANDLE)
+		{
+			WARNING_MSG("Session object handle %lu (session %lu) was not released before "
+				"HandleManager destruction; releasing now",
+				kv.first, kv.second.hSession);
+			kv.second.object->release();
+		}
+	}
 
 	MutexFactory::i()->recycleMutex(handlesMutex);
 }
 
-CK_SESSION_HANDLE HandleManager::addSession(CK_SLOT_ID slotID, CK_VOID_PTR session)
+CK_SESSION_HANDLE HandleManager::addSession(CK_SLOT_ID slotID, Session *session)
 {
 	MutexLocker lock(handlesMutex);
 
 	Handle h( CKH_SESSION, slotID );
-	h.object = session;
+	h.session = session;
 	handles[++handleCounter] = h;
 	slotHandles[slotID].insert(handleCounter);
 	slotSessionCount[slotID]++;
 	return (CK_SESSION_HANDLE)handleCounter;
 }
 
-CK_VOID_PTR HandleManager::getSession(const CK_SESSION_HANDLE hSession)
+Session *HandleManager::getSession(const CK_SESSION_HANDLE hSession)
 {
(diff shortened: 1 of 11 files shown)


src/lib/object_store/test/SessionObjectStoreTests.cpp
```diff
@@ -181,115 +181,99 @@ void SessionObjectStoreTests::testMultiSession()
...
-        objects.clear();
+	objects.clear();
 	store->getObjects(objects);
+	memset(present, 0, sizeof(present));
+	for (std::set<OSObject*>::iterator i = objects.begin(); i != objects.end(); ++i)
+	{
+		for (int j = 0; j < 5; ++j)
+		{
+			if ((*i)->getAttribute(CKA_ID).getByteStringValue() == id[j])
+			{
```
antoinelochet:
```suggestion
   for (auto* obj : objects)
	{
		for (int j = 0; j < 5; ++j)
		{
			if (obj->getAttribute(CKA_ID).getByteStringValue() == id[j])
			{
```

src/lib/handle_mgr/test/HandleManagerTests.cpp
```diff
@@ -103,76 +135,68 @@ void HandleManagerTests::testHandleManager()
...
 
-	// Logout, now private objects should be gone.
+	// Logout: private objects (both token and session) should be gone.
 	handleManager->tokenLoggedOut(slotID);
-	CPPUNIT_ASSERT(object == handleManager->getObject(hObject));
-	CPPUNIT_ASSERT(NULL == handleManager->getObject(hObject2)); // should still be private and removed.
-	CPPUNIT_ASSERT(NULL == handleManager->getObject(hObject3));
-	CPPUNIT_ASSERT(NULL == handleManager->getObject(hObject4));
-	CPPUNIT_ASSERT(object5 == handleManager->getObject(hObject5));
+	CPPUNIT_ASSERT(object  == handleManager->getObject(hObject));   // public token
```
antoinelochet: Formatting seems off

src/lib/handle_mgr/HandleManager.cpp
```diff
@@ -56,40 +56,46 @@ HandleManager::HandleManager()
 // Destructor
 HandleManager::~HandleManager()
 {
+	// Safety net to release any session-object handles that weren't explicitly cleaned up.
+    // In normal C_Finalize flow this should be empty.
```
antoinelochet: Would it be interesting to have a log message is such cases as it is the case when secrets are not recycled ?

antoinelochet (approved):

src/lib/handle_mgr/HandleManager.cpp
bukka: Good point. Added a warning for that case.

PR #874, opened 2026-05-15
Added support for ML-KEM

antoinelochet:
https://github.com/softhsm/SoftHSMv2/pull/870 must be merged first

## Summary by CodeRabbit

* **New Features**
  * ML-KEM added: key types, on-token storage, keygen, import, encapsulate/decapsulate and new PKCS#11 APIs.

* **Build & Configuration**
  * New ML-KEM/ML-DSA feature flags and detection in configure/CMake with documented options.

* **Tests**
  * New unit and integration tests plus import-key scripts for ML-DSA/ML-KEM; test wiring added.

* **Documentation**
  * README and CMake notes updated to document ML-KEM/ML-DSA options.

* **CI & Tooling**
  * Expanded Windows CI matrix, dynamic Windows dependency generation, improved CI scripts; .gitignore updated.

* **Bug Fixes**
  * Safer log buffering and various robustness fixes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/softhsm/SoftHSMv2/pull/874)

PR #881, opened 2026-06-10
Use default OpenSSL provider for internal ML-DSA key reconstruction

olszomal:
### Current Behavior:
Relying on implicit provider selection can make `ML-DSA` public/private key reconstruction fail, causing operations such as `C_Sign` to return `CKR_GENERAL_ERROR`.

### Scope of Changes:
This change explicitly uses `provider=default` when creating internal `ML-DSA` key contexts.

### Testing:
Verified `ML-DSA` signing with SoftHSM, OpenSSL 3.6.2, and a development branch of the `pkcs11prov` provider.

Before this change, signing failed because SoftHSM could not obtain the internal OpenSSL private key. After the change, signing succeeds.

### Related Pull Requests:
#823, #862, #874

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined internal cryptographic key handling for ML-DSA algorithms by explicitly configuring default provider initialization during the key reconstruction process. This ensures consistent and reliable behavior for both private and public key operations, improving system stability and compatibility with OpenSSL cryptographic frameworks.

### DIFF
--- a/src/lib/crypto/OSSLMLDSAPrivateKey.cpp
+++ b/src/lib/crypto/OSSLMLDSAPrivateKey.cpp
@@ -203,7 +203,8 @@ void OSSLMLDSAPrivateKey::createOSSLKey()
 
 	*p = OSSL_PARAM_construct_end();
 
-	ctx = EVP_PKEY_CTX_new_from_name(NULL, name, NULL);
+	// Use the default provider for internal ML-DSA key reconstruction.
+	ctx = EVP_PKEY_CTX_new_from_name(NULL, name, "provider=default");
 	if (ctx == NULL) {
 		ERROR_MSG("Could not create context");
 		return;
--- a/src/lib/crypto/OSSLMLDSAPublicKey.cpp
+++ b/src/lib/crypto/OSSLMLDSAPublicKey.cpp
@@ -123,7 +123,8 @@ void OSSLMLDSAPublicKey::createOSSLKey()
 
 	*p = OSSL_PARAM_construct_end();
 
-	ctx = EVP_PKEY_CTX_new_from_name(NULL, name, NULL);
+	// Use the default provider for internal ML-DSA key reconstruction
+	ctx = EVP_PKEY_CTX_new_from_name(NULL, name, "provider=default");
 	if (ctx == NULL) {
 		ERROR_MSG("Could not create context");
 		return;


bukka:
Can you elaborate when this is going to fail?

The preference should really be to create a custom provider and use that instead. That would resolve some other issues as well.

olszomal:
This fails when SoftHSM is used through another OpenSSL provider that also implements ML-DSA keymgmt, for example `pkcs11prov`.

SoftHSM uses OpenSSL internally as its software crypto backend. During ML-DSA signing it reconstructs an internal OpenSSL EVP_PKEY from the stored key material. Currently this uses:
```
EVP_PKEY_CTX_new_from_name(NULL, name, NULL);
```
With `propq == NULL`, the ML-DSA keymgmt fetch may resolve to one of the providers loaded by the application instead of OpenSSL's default provider. In my test case it resolves to `pkcs11prov`, and then `EVP_PKEY_fromdata()` fails.

I added temporary debug around the fetch:
``` c
EVP_KEYMGMT *keymgmt = EVP_KEYMGMT_fetch(NULL, name, NULL);
const OSSL_PROVIDER *prov = EVP_KEYMGMT_get0_provider(keymgmt);

printf("Implicit ML-DSA keymgmt provider: %s\n", OSSL_PROVIDER_get0_name(prov));
EVP_KEYMGMT_free(keymgmt);

ctx = EVP_PKEY_CTX_new_from_name(NULL, name, NULL);
...
rv = EVP_PKEY_fromdata(ctx, &pkey, selection, params);
if (rv <= 0) {
	ERROR_MSG("Could not EVP_PKEY_fromdata:%d", rv);
	printf("Could not EVP_PKEY_fromdata:%d\n", rv);
	EVP_PKEY_CTX_free(ctx);
	return;
}
```
Reproducer:
```
LD_LIBRARY_PATH=/opt/openssl-3.6.2/lib64 \
PKCS11_MODULE_PATH=/opt/softhsm2/lib/softhsm/libsofthsm2.so \
SOFTHSM2_CONF=/home/mo/libp11/tests/output.91001/softhsm-testpkcs11.config \
/opt/openssl-3.6.2/bin/openssl pkeyutl \
  -provider pkcs11prov -provider default \
  -sign \
  -in /tmp/hash.bin \
  -out /tmp/signature.bin \
  -inkey "pkcs11:token=token1;object=mldsa87_key;type=private"
```
Output:
```
Found 1 private key:
   1 P  id=2233 label=libp11-keylabel
Returning last matching private key: id=2233 label=libp11-keylabel
p11_pkey.c:376 pkcs11_sign_with_mechanism() CKM_ML_DSA sig=0xb59ed0 *siglen=4627 tbs=0xb2efd0 tbslen=32
Implicit ML-DSA keymgmt provider: pkcs11prov
Could not EVP_PKEY_fromdata:0
p11_pkey.c:412 C_Sign rv=5
Public Key operation error
0012D600C87F0000:error:030000EA:digital envelope routines:EVP_DigestSign:provider signature failure:crypto/evp/m_sigver.c:649:PKCS11 digest_sign:PKCS#11 signature functions

```
So the failure chain is:
- `pkcs11prov` calls SoftHSM through PKCS#11.
- SoftHSM tries to reconstruct its internal ML-DSA OpenSSL key.
- The implicit keymgmt fetch resolves to `pkcs11prov`.
- EVP_PKEY_fromdata() fails.
- SoftHSM cannot obtain the internal OpenSSL private key and returns CKR_GENERAL_ERROR.

In this code path, SoftHSM is not trying to delegate ML-DSA key reconstruction to an application-selected provider. It is reconstructing its own internal software key for its OpenSSL crypto backend. For that reason I think it should explicitly use OpenSSL's default provider here, rather than relying on implicit provider selection from the application's provider configuration.

petrovr:
This should be resolved by OpenSSL configuration.
Providers designed for "external" keys  cannot be used for "generic" functionality.

PR #833, opened 2025-12-24
Refactor rsa/add modern hashs in OAEP encrypt-decrypt

noiser7:
Refactor RSA for OpenSSL 3.0
Add new SHA hashes and MGF for RSA OAEP encryption/decryption
Add label support for OAEP encryption

## Summary by CodeRabbit

* **New Features**
  * Full RSA OAEP support: OAEP parameters (hash, MGF, label/source) are built and forwarded across encryption, decryption, wrap and unwrap flows.

* **Improvements**
  * Asymmetric operations now accept optional auxiliary parameter payloads for OAEP/context data.
  * RSA backend modernized and memory handling improved.
  * OAEP validation relaxed for broader interoperability.

* **Tests**
  * Expanded RSA/OAEP tests covering multiple digests, MGF variants, labels, and larger keys.

PR #882, opened 2026-06-11
Add legacy RSA fallback for Engine-based applications

olszomal:
### Current Behavior:
With OpenSSL 3.x, RSA key reconstruction uses the provider-based EVP_PKEY_fromdata() path.

In Engine-based applications, a suitable provider context may not always be available for this operation. In that case, EVP_PKEY_fromdata_init() can fail, which prevents SoftHSM from reconstructing the internal RSA key object.

### Scope of Changes:
This change explicitly uses the OpenSSL default provider for internal RSA key reconstruction.

If the provider-based RSA reconstruction cannot be initialized, the code falls back to the legacy RSA API while it is still available. This fallback is intended for Engine-based applications and preserves compatibility with OpenSSL 3.x.

The fallback is only used when the provider key creation context cannot be initialized. Failures from EVP_PKEY_fromdata() itself are still treated as RSA key reconstruction errors.

### Testing:
Verified RSA signing with SoftHSM, OpenSSL 3.6.2, and the libp11 Engine.

Before this change, RSA signing through the Engine failed because SoftHSM could not initialize the provider-based RSA key reconstruction path. After this change, SoftHSM falls back to the legacy RSA API and signing succeeds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced RSA cryptographic key handling to improve compatibility with OpenSSL 3.x and 4.x versions.
  * Improved error handling and fallback mechanisms for more reliable RSA key reconstruction operations.
  * Refined diagnostic handling and error state management for better stability across OpenSSL versions.

### DIFF
--- a/src/lib/crypto/OSSLRSAPrivateKey.cpp
+++ b/src/lib/crypto/OSSLRSAPrivateKey.cpp
@@ -39,9 +39,9 @@
 #include <openssl/x509.h>
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 #include <openssl/param_build.h>
-#else
-#include <openssl/rsa.h>
+#include <openssl/err.h>
 #endif
+#include <openssl/rsa.h>
 #ifdef WITH_FIPS
 #include <openssl/fips.h>
 #endif
@@ -354,7 +354,9 @@ void OSSLRSAPrivateKey::createOSSLKey()
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
 	OSSL_PARAM_BLD* param_bld = OSSL_PARAM_BLD_new();
 	OSSL_PARAM* params = NULL;
+	EVP_PKEY_CTX* ctx = NULL;
 	bool bBuildErr = false;
+
 	if ((param_bld == NULL) ||
 		(bn_n == NULL) ||
 		(bn_e == NULL) ||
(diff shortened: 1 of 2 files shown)


bukka:
I don't really like this as we we would apply this kind of logic to all algs - note that this is likely already broken for DH. I'm not keen on this sort of deprecation suppression as it's quite ugly hack.

What I would be willing to accept is to completely separate old api to the new backend that would be primarily for LibreSSL could support 1.1.1 and legacy engine code in 3.x. So if you want to go with this, you will need to copy all algs so we will have some duplicate code for a while but it would make the separation easier when we migrate other algs. So I think that's the path.

olszomal:
Thanks for the feedback.

I agree that splitting the legacy OpenSSL API backend from the provider-based backend would be cleaner long term, but that is larger architectural work and outside the scope of this PR.

This PR fixes a specific regression from #833: RSA signing through the libp11 Engine worked before the RSA key reconstruction was moved to the `EVP_PKEY_fromdata()` path. With OpenSSL 3.x, that provider-based path cannot be initialized in the Engine setup, so RSA signing fails.

The legacy RSA code already existed. This PR only keeps it usable for the OpenSSL 3.x Engine compatibility case.

petrovr:
This path explicitly specify a provider. Such model is not acceptable for build with OpenSSL 3.0.
What if Linux is installed in FIPS mode?

PR #884, opened 2026-06-20
Fix size_t-to-int truncation in the OpenSSL RNG backend

mingulov:
RAND_bytes() and RAND_seed() take an int length, so on 64-bit platforms any request above INT_MAX was silently truncated: a >2 GiB C_GenerateRandom() left most of the buffer unwritten while still reporting success, and C_SeedRandom() used only a truncated prefix of the seed.

Both now clamp each call to INT_MAX and loop, so the whole length is honored. RAND_seed() accumulates entropy across calls, so feeding it in pieces is fine. The Botan backend takes size_t and was already fine.

Demo without this fix (on 64-bit platforms with enough RAM):
```
  CK_ULONG    size = 0x100000008UL;                 /* 4 GiB + 8 bytes */
  CK_BYTE_PTR buf  = (CK_BYTE_PTR)calloc(size, 1);   /* zero-filled */
  CK_RV       rv   = C_GenerateRandom(s, buf, size);
  /* rv == CKR_OK, yet only buf[0..7] are randomised; buf[8..] stays 0 (4 GiB of zeros) */
```

## Release Notes

* **Bug Fixes**
  * Improved the cryptographic random number generator to properly handle requests for very large amounts of random data, preventing potential truncation issues that could occur with oversized requests.

### DIFF
--- a/src/lib/crypto/OSSLRNG.cpp
+++ b/src/lib/crypto/OSSLRNG.cpp
@@ -33,20 +33,44 @@
 #include "config.h"
 #include "OSSLRNG.h"
 #include <openssl/rand.h>
+#include <limits.h>
+
+// RAND_bytes()/RAND_seed() take an int length, so a size_t over INT_MAX would
+// be truncated. Clamp each call and loop so the whole request is covered.
+static int clampToInt(size_t len)
+{
+	return len > (size_t) INT_MAX ? INT_MAX : (int) len;
+}
 
 // Generate random data
 bool OSSLRNG::generateRandom(ByteString& data, const size_t len)
 {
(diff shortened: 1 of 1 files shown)


mingulov:
An additional analysis: C_GenerateRandom does not leak internal memory, those zeroes are from ByteString::wipe.
So remote access methods like different pkcs11 proxies do not expose stale heap etc.

No test included as it requires a lot of resources - softhsm2 uses an additional own buffer, so for the minimum 4 GB test - 8 GB will be needed. 
Found while developing pkcs11-check and pkcs11-proxy-ng.

bukka (approved):

PR #891, opened 2026-06-27
fixed wrong usage of VirtualFree (win32)

frankmorgner:
SecureAlloc should have never even compiled on win32. First, the API usage is wrong - VirtualFree requires a size parameter (0 in case of MEM_RELEASE). Second, the deallocator used a non existing variable name as reference.

This change fixes the usage, but the change itself is untested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows memory cleanup for secure non-paged allocations, including correct handling of allocation-failure cleanup.
  * Updated secure memory deallocation on Windows to better match the expected `VirtualFree` call pattern, reducing the risk of cleanup issues.

### DIFF
--- a/src/lib/data_mgr/SecureAllocator.h
+++ b/src/lib/data_mgr/SecureAllocator.h
@@ -122,7 +122,7 @@ public:
 #ifndef _WIN32
 			free(r);
 #else
-			VirtualFree((const void*) r, MEM_RELEASE);
+			VirtualFree((const void*) r, 0, MEM_RELEASE);
 #endif
 
 			return NULL;
@@ -173,7 +173,7 @@ public:
 #ifndef _WIN32
 		free(p);
 #else
-		VirtualFree((const void*) r, MEM_RELEASE);
+		VirtualFree((const void*) p, 0, MEM_RELEASE);
 #endif
 #else
 		// Release the memory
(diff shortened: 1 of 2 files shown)


bukka (changes requested):
The CodeRabbit comments seem correct, please fix them.

frankmorgner:
> The CodeRabbit comments seem correct, please fix them.

First, I don't think so. Which comment do you mean in particular?

Second, your code obviously was *never* tested on Windows. It wasn't even compiled on Windows (otherwise the compiler would have complained). So you may want to delete it entirely... Without a meaningful test or any actual user, you are only carrying dead weight.

frankmorgner:
The implementation in SecureAllocator.h comes from 1b80d8356927c81b7531641fb90ad9e03744fb00 and it wasn't changed significantly since then. Note that the text in the commit reads "tentative port to WIN32", where the *tentative* seemingly should be taken seriously.

bukka:
Ah, it actually just suggests that there are two additional problems which seems correct but the whole thing is actually dead code because SENSITIVE_NON_PAGED is always defined.

We build SoftHSMv2 on Windows on CI as you can see but this is just dead code...

bukka:
Your change is correct per se but need to think if it's actually even worth keeping.

frankmorgner:
Me personally, I would review the code needed (or not needed) - but that would be something I am leaving to you as I am not too familiar with the code base. If YOU put this on your to do list (or into a GH issue), then I would close this PR.

bukka (approved):
As this fixes part of the problem, I think it's fine to merge it. I need to make a note to see if this whole code is something we want to keep.

PR #892, opened 2026-07-06
Fixed EDDSA context/pre-hash support

antoinelochet:
Fix for https://github.com/softhsm/SoftHSMv2/issues/873

## New Features
- Added support for EdDSA context data and pre-hashed signing and verification.
- Expanded Ed25519 and Ed448 support, including context-aware and pre-hashed variants.

## Bug Fixes
- Improved validation of invalid, incomplete, or mismatched EdDSA parameters.
- Enhanced compatibility across supported cryptographic backends and versions.
- Improved error reporting for EdDSA operation failures.

## Tests
- Added coverage for context and pre-hash scenarios.
- Added known-vector tests and negative verification cases for Ed25519 and Ed448.

### DIFF
--- a/src/lib/crypto/BotanEDDSA.cpp
+++ b/src/lib/crypto/BotanEDDSA.cpp
@@ -38,6 +38,7 @@
 #include "CryptoFactory.h"
 #include "BotanCryptoFactory.h"
 #include "ECParameters.h"
+#include "EDDSAMechanismParam.h"
 #include "BotanEDKeyPair.h"
 #include "BotanUtil.h"
 #include <algorithm>
@@ -62,22 +63,56 @@ BotanEDDSA::~BotanEDDSA()
 	delete verifier;
 }
 
+// Select the Botan EMSA for the requested EdDSA instance
+bool BotanEDDSA::selectEmsa(std::string& emsa, size_t orderLength, const MechanismParam* mechanismParam)
+{
+	emsa = "Pure";
+
+	if (mechanismParam == NULL || !mechanismParam->isOfType(EDDSAMechanismParam::type))
+	{
+		return true;
+	}
+
+	const EDDSAMechanismParam* eddsaParam = (const EDDSAMechanismParam*) mechanismParam;
+
+	if (eddsaParam->contextData.size() > 0)
+	{
+		ERROR_MSG("EDDSA: Context data is not supported in the Botan 2 EDDSA implementation");
+		return false;
+	}
+
+	if (!eddsaParam->flag)
+	{
+		return true;
+	}
+
+	// Ed25519: orderLength = 32, Ed448: orderLength = 57
+	if (orderLength == 32)
+	{
+		emsa = "Ed25519ph";
+		return true;
+	}
+
+	if (orderLength == 57)
+	{
+		ERROR_MSG("EDDSA: Ed448 is not supported in the Botan 2 EDDSA implementation");
+		return false;
+	}
+
+	ERROR_MSG("EDDSA: Unknown EdDSA key size: %lu", (unsigned long) orderLength);
+	return false;
+}
+
 // Signing functions
 bool BotanEDDSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
 		      ByteString& signature, const AsymMech::Type mechanism,
-		      const MechanismParam* /* mechanismParam */)
+		      const MechanismParam* mechanismParam)
 {
-	std::string emsa;
-
-	if (mechanism == AsymMech::EDDSA)
+	if (mechanism != AsymMech::EDDSA)
 	{
-		emsa = "Pure";
-	}
-        else
-        {
 		ERROR_MSG("Invalid mechanism supplied (%i)", mechanism);
 		return false;
-        }
+	}
 
 	// Check if the private key is the right type
 	if (!privateKey->isOfType(BotanEDPrivateKey::type))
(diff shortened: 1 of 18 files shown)


antoinelochet:
@coderabbitai review

src/lib/crypto/OSSLEDDSA.cpp
```diff
@@ -69,31 +74,100 @@ bool OSSLEDDSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
...
+			{
+				instance = "Ed448ph";
+			}
+		}
+		else
+		{
+			ERROR_MSG("EDDSA: Unknown EdDSA key size: %zu", pk->getOrderLength());
+			EVP_MD_CTX_free(ctx);
+			return false;
+		}
```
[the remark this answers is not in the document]
antoinelochet: @coderabbitai Propose a patch

src/lib/crypto/BotanEDDSA.cpp
```diff
@@ -195,6 +229,39 @@ bool BotanEDDSA::verify(PublicKey* publicKey, const ByteString& originalData,
...
+			return false;
+		}
+
+		if (pk->getOrderLength() == 32)
+		{
+			emsa = "Ed25519ph";
+		}
+		else
+		{
+			emsa = "Ed448ph";
```
Jakuje: same as in the openssl case, there should be explicit check for the order length 57B and fail in other cases as other "key sizes" are not defined by eddsa

src/lib/test/SignVerifyTests.cpp
```diff
@@ -844,6 +844,189 @@ void SignVerifyTests::testEdSignVerify(const char* curve)
...
+	// Login USER into the sessions so we can create a private objects
+	rv = CRYPTOKI_F_PTR( C_Login(hSessionRO,CKU_USER,m_userPin1,m_userPin1Length) );
+	CPPUNIT_ASSERT(rv==CKR_OK);
+
+	CK_OBJECT_HANDLE hPuk = CK_INVALID_HANDLE;
+	CK_OBJECT_HANDLE hPrk = CK_INVALID_HANDLE;
+
+	// Test EdDSA signature with context data
+	// Create EdDSA parameters with context
+	CK_BYTE contextData[] = { 0x4a, 0x65, 0x73, 0x73, 0x69, 0x63, 0x61 }; // "Jessica"
```
Jakuje: Is this from some known test vector?

antoinelochet: Nope. Just a dumb test because I can not make it work right now.
Mecanism parameters are not passed on to C_SignInit and I do not know why yet.

antoinelochet:
@coderabbitai review

src/lib/test/SignVerifyTests.cpp
```diff
@@ -784,6 +784,194 @@ void SignVerifyTests::testEdSignVerify(const char* curve)
...
+	CPPUNIT_ASSERT(rv == CKR_OK);
+	signVerifySingle(CKM_EDDSA, hSessionRO, hPuk, hPrk, &eddsaParams, sizeof(CK_EDDSA_PARAMS));
+
+	// Test with different context data
+	CK_BYTE anotherContext[] = { 0x13, 0xe8, 0x06, 0x38, 0x7a, 0xdc, 0x2d, 0x73, 0x3c, 0x85, 0xa9, 0xd0, 0x81, 0x91, 0xfa, 0xa0, 0xcd, 0xeb, 0x11, 0xca, 0x4d, 0x1c, 0x2a, 0x05, 0x7c, 0x27, 0xf3, 0x6c, 0xeb, 0xc4, 0xdf, 0x88, 0x8a, 0x45, 0x6a, 0xc5, 0xc0, 0x91, 0x69, 0x31, 0x4e, 0xb0, 0x49, 0xe7, 0xdf, 0xdc, 0xf8, 0x68, 0x67, 0x21, 0xf6, 0xda, 0x13, 0x46, 0x1c, 0x57, 0x5c, 0x6e, 0x78, 0x36, 0x91, 0xc4, 0x2d, 0x09 }; //sha512("context-data-bob") for Ed25519ph
+	CK_EDDSA_PARAMS eddsaParams2 = {
+		0x01,  // phFlag = 1 (pre-hash)
+		sizeof(anotherContext),  // context_data_len
+		anotherContext  // context_data
+	};
```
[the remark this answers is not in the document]
antoinelochet: @coderabbitai Suggest a fix

src/lib/crypto/test/EDDSATests.cpp
```diff
@@ -43,6 +43,33 @@
...
+#if defined(WITH_OPENSSL) && OPENSSL_VERSION_NUMBER < 0x30200000L
+	#define EDDSA_HAVE_CONTEXT	0
+	#define EDDSA_HAVE_PREHASH	0
+#elif defined(WITH_BOTAN)
+	#define EDDSA_HAVE_CONTEXT	0
+	#define EDDSA_HAVE_PREHASH	1
+#else
+	#define EDDSA_HAVE_CONTEXT	1
+	#define EDDSA_HAVE_PREHASH	1
+#endif
```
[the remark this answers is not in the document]
antoinelochet: ignore

src/lib/crypto/BotanEDDSA.cpp
```diff
@@ -62,22 +63,56 @@ BotanEDDSA::~BotanEDDSA()
...
+
+	if (!eddsaParam->flag)
+	{
+		return true;
+	}
+
+	// Ed25519: orderLength = 32, Ed448: orderLength = 57
+	if (orderLength == 32)
+	{
+		emsa = "Ed25519ph";
```
bukka, line 92: Looks like this was introduced in Botan 2.11 so I think we should just bump minimum to Botan . I don't think anyone is using older than 2.12 (RHEL 8 version) so I would just go for that for now.

src/lib/crypto/OSSLEDDSA.cpp
```diff
@@ -69,11 +127,28 @@ bool OSSLEDDSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
...
+		ByteString contextData;
+#endif
+	if (mechanismParam != NULL && mechanismParam->isOfType(EDDSAMechanismParam::type))
+	{
+#if OPENSSL_VERSION_NUMBER >= 0x30200000L
+			EDDSAMechanismParam* eddsaParam = (EDDSAMechanismParam*) mechanismParam;
+			preHash = eddsaParam->flag;
+			contextData = eddsaParam->contextData;
+#else
+			ERROR_MSG("Context data is not supported in OpenSSL < 3.2");
```
bukka: This means that `{phFlag = CK_FALSE, ulContextDataLen = 0, pContextData = NULL}` will not work which it should even on lower versions. Botan seems to already handle it.

bukka (changes requested):

PR #894, opened 2026-07-22
Add SHA3 mechanism support

### LINKED ISSUE #603: Add support for SHA3 digest mechanisms
Enhancement: Add support for SHA3 family of hashing algorithms

To support the following mechanisms:
CKM_SHA3_*
CKM_ECDSA_SHA3_*

Please consider adding SHA3 support.

AlexLaroche:
## Summary

Implements the full PKCS#11 v3.2 SHA3 (FIPS 202) mechanism family for both the OpenSSL and Botan crypto backends:

- Plain digest (`CKM_SHA3_224/256/384/512`)
- HMAC (`CKM_SHA3_*_HMAC`)
- RSA combined signing (`CKM_SHA3_*_RSA_PKCS`, `CKM_SHA3_*_RSA_PKCS_PSS`), including the raw/precomputed-digest `CKM_RSA_PKCS_PSS` path
- DSA combined signing (`CKM_DSA_SHA3_*`)
- ECDSA combined signing (`CKM_ECDSA_SHA3_*`)
- SHA3 as an RSA-OAEP hash parameter

Not included: `CKM_SHA3_*_KEY_GEN` and `CKM_SHA3_*_HMAC_GENERAL` — neither is implemented for any existing hash family today, so neither is added here either.

Closes #603.

## Test plan

- [x] `HashTests`: FIPS 202 known-answer vectors for SHA3-224/256/384/512, both backends
- [x] `DigestTests`: digest round-trip for all four sizes
- [x] `SignVerifyTests`: HMAC-SHA3 round-trip; RSA-PKCS and RSA-PKCS-PSS round-trip (named and raw/precomputed-digest); ECDSA round-trip across P-256/P-384/P-521; DSA wrong-key-type negative cases extended (no positive DSA+hash round-trip exists for any size today, so none was added here either)
- [x] `AsymEncryptDecryptTests`: RSA-OAEP round-trip with a SHA3 hash parameter, with and without a label
- [x] Full test suite green on both the OpenSSL and Botan crypto backends

## Summary by CodeRabbit

* **New Features**
  * Added SHA-3 support for hashing, HMAC, and SHA-3–based PKCS#11 operations including RSA (PKCS, RSA-PSS), RSA-OAEP, DSA, and ECDSA.
  * Added SHA3-224/256/384/512 mechanism recognition (including SHA3-based MGF1) and improved mechanism info reporting and secret-key check handling.
* **Tests**
  * Expanded automated coverage for SHA-3 digests, HMAC, sign/verify (RSA/DSA/ECDSA), and RSA-OAEP.
* **Documentation**
  * Added `--enable-sha3` configure option.


src/lib/crypto/OSSLRSA.cpp
```diff
@@ -177,6 +177,22 @@ bool OSSLRSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
...
+				allowedLen = 32;
+				break;
+			case HashAlgo::SHA3_384:
+				hash = EVP_sha3_384();
+				allowedLen = 48;
+				break;
+			case HashAlgo::SHA3_512:
+				hash = EVP_sha3_512();
+				allowedLen = 64;
+				break;
```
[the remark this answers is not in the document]
AlexLaroche: Resolved differently than suggested here — rather than raising the OpenSSL/LibreSSL minimum, the `EVP_sha3_*` call sites (`OSSLRSA.cpp`) and every other SHA3 touch point (`OSSLSHA3_*`/`BotanSHA3_*` classes, HMAC-SHA3, factory dispatch, `SoftHSM.cpp` mechanism registration/dispatch) are now gated behind `WITH_SHA3`, only defined when `--enable-sha3` (default detect) confirms real `EVP_sha3_*` support at configure time — mirroring the existing `--enable-ecc`/`--enable-eddsa` pattern. OpenSSL's floor stays at 1.0.0/1.0.1, unchanged. See b349775.

@coderabbitai resolve

src/lib/crypto/BotanRSA.cpp
```diff
@@ -333,6 +345,122 @@ bool BotanRSA::signInit(PrivateKey* privateKey, const AsymMech::Type mechanism,
...
+			if ((pssParam->hashAlg != HashAlgo::SHA3_224)||
+			    (pssParam->mgfAlg != AsymRSAMGF::MGF1_SHA3_224))
+			{
+				ERROR_MSG("Invalid RSA PSS mechanism parameters supplied");
+				ByteString dummy;
+				AsymmetricAlgorithm::signFinal(dummy);
+				return false;
+			}
+			sLen = pssParam->sLen;
+			if (sLen > ((privateKey->getBitLength()+6)/8-2-28))
```
antoinelochet: Could you factor this check in a dedicated method to avoid repetition below ?

AlexLaroche: Factored the check into `BotanRSA::checkPssParams()` in 663cf20 — it validates the mechanism-param type, hash/MGF match, and salt-length bound, and is now called from all 10 PSS cases in both `signInit` and `verifyInit` instead of being repeated inline.

antoinelochet: I just saw (sorry) that about the same could be done in `OSSLRSA`.

AlexLaroche: Good catch, no worries. Applied the same factoring to `OSSLRSA::checkPssParams()` in OSSLRSA.cpp — same signature, same 10 call sites deduplicated in both signInit and verifyInit. Verified with both crypto backends (Botan and OpenSSL): full build clean, cryptotest and p11test (including the SHA3 PSS sign/verify paths) all pass.

src/lib/crypto/OSSLRSA.cpp
```diff
@@ -337,6 +337,34 @@ bool OSSLRSA::sign(PrivateKey* privateKey, const ByteString& dataToSign,
 	}
 }
 
+// Validate an RSA-PSS mechanism parameter against the expected hash/MGF algorithm and
+// check that the requested salt length fits the key size. Returns the salt length in sLen
+// on success; logs an ERROR_MSG and returns false otherwise.
+bool OSSLRSA::checkPssParams(const MechanismParam* mechanismParam, HashAlgo::Type hashAlg,
```
antoinelochet: `OSSLRSA::checkPssParams` and `BotanRSA::checkPssParams` are identical.
I think creating a `RSAUtil` class, as it is done for ML-DSA and ML-KEM and put that method in this newly-created class, could be a solution to this problem.

AlexLaroche: Makes sense. Created `RSAUtil` (RSAUtil.h/.cpp) following the MLDSAUtil/MLKEMUtil pattern, moved `checkPssParams()` there, and both `BotanRSA` and `OSSLRSA` now call `RSAUtil::checkPssParams(...)` instead of keeping their own copy. Wired into both Makefile.am and CMakeLists.txt. Verified with both backends (Botan and OpenSSL): full build clean, cryptotest and p11test (SHA3 PSS sign/verify included) all pass. Pushed in 2563983.

antoinelochet: Very nice. Thank you.

antoinelochet:
I think this one is ready for merging.

src/lib/crypto/BotanSHA3_224.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE
+ * GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+ * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN
+ * IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
bukka: For new files, we just use this.
```suggestion
 * Copyright (c) 2026 SoftHSMv2 contributors
 *
 * SPDX-License-Identifier: BSD-2-Clause
 */
```

bukka: Same applies to all other new files.

config.h.in.cmake
```diff
@@ -157,6 +157,9 @@
 /* Compile with ML-KEM support */
 #cmakedefine WITH_ML_KEM @WITH_ML_KEM@
 
+/* Compile with SHA3 support */
+#cmakedefine WITH_SHA3 @WITH_SHA3@
```
bukka, line 161: how does this get enabled in cmake?

bukka (changes requested):
It looks good, just some minor things.

src/lib/crypto/BotanSHA3_224.cpp
AlexLaroche: Fixed in accc4d7.

AlexLaroche: Applied to all 16 new files (BotanSHA3_*/OSSLSHA3_* .cpp/.h) in accc4d7.

config.h.in.cmake
AlexLaroche: Good catch — it didn't. config.h.in.cmake referenced @WITH_SHA3@ but nothing in the CMake side ever set that variable, so any CMake build was silently compiling out SHA3 support entirely (RSA-SHA3-PKCS, RSA-SHA3-PSS, HMAC-SHA3, digests, everything gated behind WITH_SHA3).

Fixed in 9b8b40c by mirroring the autotools logic in cmake/modules/CompilerOptions.cmake: Botan always has SHA3 (matches the "no version detection is needed" comment in acx_crypto_backend.m4) so it's set unconditionally there; OpenSSL gets a try_run sanity check (new cmake/modules/tests/test_openssl_sha3.c, same EVP_sha3_* probe as acx_openssl_sha3.m4) requiring OpenSSL >= 1.1.1 / LibreSSL >= 3.8.0.

Verified with an actual CMake configure+build (OpenSSL backend): WITH_SHA3 lands in the generated config.h, full build is clean, and cryptotest/p11test (81 tests, including SHA3 PSS sign/verify) all pass. The Botan+CMake side reports "Botan: SHA3 is supported" and WITH_SHA3 is set correctly too, but I couldn't get a full Botan CMake build to link — that's blocked by an unrelated pre-existing bug (src/lib/crypto/CMakeLists.txt:171 unconditionally appends MSVC-style /wd* flags for any WITH_BOTAN build regardless of compiler, from 2021, commit 8fd89ec3). Happy to file that separately if useful, didn't want to fold an unrelated fix into this PR.

bukka (approved):

AlexLaroche:
@bukka, @antoinelochet : Is anything missing for it to be merged?

antoinelochet:
> @bukka, @antoinelochet : Is anything missing for it to be merged?

I do not have the rights to merge.
ping @bukka

bukka:
Sorry, somehow forgot to merge it before and then had holiday and busy with other things.

PR #895, opened 2026-08-01
silent automake build

petrovr:
## Summary by CodeRabbit

* **Build Experience**
  * Enabled quieter build output by default when supported.
  * Added options to restore detailed build messages or override the default verbosity when needed.


petrovr:
pushed again without unrelated changes for #864

